PR #16, opened 2020-05-18
Feature/readme getting started

mlange-42:
### Features

* Create config toml files with comments on all options

### Documentation

* Moved feature documentation to docs/Features.md
* Fine-tuning of all documentation

### Fixes

* Code cleanup


README.md
```diff
@@ -35,14 +36,22 @@ yarner create README.md
...
+basic features. For details, read the following sections.
 
 To "compile" the project (extract code and create documentation),
 simply run:
+
 ```
 yarner
 ```
 
+This creates two sub-directories, one containing the extracted code,
```
adamreichold, line 52: Maybe mention that both then need to be built using their "usual" build systems?

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+Now, to say the default "Hello" to many people:
+
+```rust
+==> Say @{} to @{Jim}.
+==> Say @{} to @{Tom}.
+==> Say @{} to @{Angela}.
+```
+~~~
+
+These features allows for more flexibility when writing macros, as well as possibly making the intent
```
adamreichold, line 104: This could be extend by a sentence that this should not be abused and the target languages mechanisms of abstraction should be preferred.

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+clearer.
+
+## Extracted comments
+
+By default, the comment extraction sequence is set to `//`, purely for familiarity. Any text after
+(and including) this sequence is extracted from the code block, and not rendered to the tangled
+source code. Note that, since the comments are removed completely when compiling, they do not have
+to use the actual line comment indicator from you programming language. In fact, it may be better to
```
adamreichold, line 112: I find that questionable, as the generated source code is not intended to be human-readable, so why should it contain comments? The document itself should do all the commentary necessary.

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+choose a sequence that is *not* the regular comment indicator so that you can still have comments in
+your tangled code output.
+
+Now that these comments are extracted, it is possible to handle them differently in the weaved
```
adamreichold, line 116: Also here, the question seems to be why not mark these comments up in the actual document because that is where the human-readable presentation is supposed to happen.

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+
+See the HTML example for an example of one way to render the extracted comments.
+
+## Named entrypoints
+
+By default, the entrypoint of the program is always the unnamed code block. However, this limits the
+output of one input file to always be the same source code. It also means that you can't have a name
+on the entrypoint in the documentation, which could be useful.
+
+To get around this, an entrypoint name can be passed to Yarner on the command line. Then, instead
```
adamreichold, line 130: Would it make sense to make the entry point name configurable via the TOML file so that I do not need to pass it on the command line for each invocation?

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+```md
+@{{[src/main.rs](src/main.rs.md)}}
+```
+which would render as a link in the sources file. Or simply:
+```md
+@{{src/main.rs.md}}
+```
+which would not render as a link.
+
+Unnamed entrypoints are renamed to file entrypoints during transclusion.
```
adamreichold, line 167: I think these last two paragraphs would benefit from a worked example now that the space limitation is gone.

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+fn say_hello() {
+    println!("Hello Literate Programmer!");
+}
+```
+~~~
+
+## File Transclusions
+
+Yarner supports file transclusion to allow for more structured or modular project sources.
+*This feature is currently only supported for Markdown.*
```
adamreichold, line 153: I wonder if this is really a limitation as described here because TeX already has mechanisms for this. Maybe this should be changed to say that it only applies to Markdown? (Same below on including relative links.)

docs/Features.md
mlange-42: Improvements to this?

> These features allows for more flexibility when writing macros, as well as possibly making the intent clearer. However, they should not be abused to replace the target languages's abstraction mechanisms. The language's mechanisms should be strictly preferred.

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+However, files are only included in the compilation if they are referenced by a *relative path*, and the file exists in that location.
+
+## Multiple languages
+
+Some documentation formats allow you to indicate the language that a code block is written in. In
+fact, it is recommended that you always include the language when you write a code block,
+particularly if multiple programming languages are used within the same document.
+
+By properly labelling all code blocks, it is then possible to write a program in multiple
+programming languages at once. Whether this is practical or not remains to be seen, but it is
```
adamreichold, line 193: Well I certainly wrote multiple programs that contained Rust, C and sometimes C++ code. Often production code is in one language while tooling for testing and deployment is written in another, C++ and Python are common combination. So I think practicality is established even though I am admittedly not sure how much this applies in a scientific context, but then again job submission scripts are Bourne shell scripts no matter what the actual modelling code is written in. (Hopefully not using Bash. I have it on good authority that a certain company from Leipzig uses a web application backend written in Bash and that this is the most probably cause if all of the cities inhabitants will someday be reborn as pieces of butter.)

docs/Features.md
```diff
@@ -0,0 +1,216 @@
...
+
+```rust
+fn main() {
+  println!("Hello world");
+}
+```
+~~~
+
+Compiling this with no language supplied with just ignore language information, so a single output
+will be generated containing both languages. However, supplying the `--language rb` flag to Yarner
```
adamreichold, line 215: Is this possible via the TOML as well as it seems like something that will not change for a given project.

docs/Features.md
mlange-42: Maybe you misunderstand that? Which means it should be expressed more clearly...

I read your comment as "//-Comments in code blocks should go into the code", as they actually do not.

Otherwise, I can only interprete it as "strip all comments out" (incl. the language's "real" comments).

docs/Features.md
adamreichold: Using `'s` is restrict to persons AFAIK and I think the last two sentences could be merged, e.g.

>  However, macros should not be abused to replace the mechanisms of abstraction provided by the target language. These are preferable as they enforce semantic in addition to purely syntactic structure.

docs/Features.md
mlange-42: Yes, this makes sense. Although it is probably rarely used. How was that with Some/None in TOML...

docs/Features.md
mlange-42: Not really. TeX has transclusions, but the "transcluded" code blocks are not accessible from the transcluding document. For links, they are not detected in TeX, and hence linked files are not "compiled".

docs/Features.md
mlange-42: This is from the old docs, but applies to individual code blocks. With file entrypoints, multiple languages are even simpler, and it may be questionable if the original way is still useful.

Different languages would mostly go to different output files, or are there exceptions?

mlange-42: Also, the featrue described here is not for "compiling" to multiple languages at the same time, while file entrypoints are.

docs/Features.md
mlange-42: This is very rarely used, I think.

docs/Features.md
mlange-42: I think removing "normal" (language) comments from the source is too harsh.

In addition, comments in the output may become useful once there is a "reverse mode". 

And again, why should humans not **want** to read the code? Maybe to check if everything works as expected. Or because they are fed up with LP, and want to continue "the normal way".

docs/Features.md
adamreichold: I am just referring to the last part about it being better to have comments in the code. I think this "advice" is a bit ill-advised as comments should be part of the document containing the code when using LP. I would probably just drop that last sentence?

docs/Features.md
adamreichold: `serde` and `toml` should handle `Option<String>` as the field type fine to support this optional item.

docs/Features.md
adamreichold: I see. You cannot pull in stuff using macros that way.

docs/Features.md
adamreichold: > Different languages would mostly go to different output files, or are there exceptions?

At least, I have never used anything else. Hence I would take up every opportunity for simplification there is and maybe remove the feature entirely or at least tone down the auxiliary discussion here.

docs/Features.md
adamreichold: As written above, if this can be considered fully redundant to named entry points, I would probably remove the feature entirely to simplify things.

adamreichold: (I also do not rarity applies here: If I do use the feature, then I probably use it each time I compile the documents of a given project. It does not change between invocations.)

docs/Features.md
adamreichold: Note that here I am referring to the Yarner-specific comments, not the target language comments. Those are put into markup-specific tags and such and this is what I am questioning: If I want to have an `<aside>` here, why not just put that into the surrounding markup instead of relying on the extraction of Yarner-specific comments?

docs/Features.md
mlange-42: For both, renaming entrypoints and "no own entrypoints"?

adamreichold: I think of them would suffice to illustrate what is happening.

docs/Features.md
mlange-42: Because this may split up code too much, or because you may somehow want to "label" particular parts of the code that you describe above or below the block?

docs/Features.md
mlange-42: Sounds reasonable!

README.md
```diff
@@ -192,73 +161,14 @@ fn say_hello() {
...
-```
-~~~
+## Configuration
 
-Compiling this with no language supplied with just ignore language information, so a single output
-will be generated containing both languages. However, supplying the `--language rb` flag to yarner
-will cause only the code blocks tagged with `rb` will be used to generate code.
+Each style supports some additional configuration, which is provided via a toml configuration file
+(default: Yarner.toml). A file with default configurations is generated through the `create` sub-command.
+See the comments in these files for details on individual settings.
```
mlange-42, line 170: @adamreichold Argh... not so easy, as the main file is filled in at runtime. Maybe I need a separate file for the config docs.

docs/Features.md
adamreichold: The `<aside>` will be an aside whether its markup is before, after or within the code block. The Yarner-specific comments will not be visible in the source code anyway. And the important thing here is the layout of the human-readable document, not the layout of its mark-up, which will be the same in both approaches. (Consider someone using e.g. LyX to edit the LaTeX document, he/she will never actually see that mark-up.) Also I cannot label stuff using comments that will be removed, right?

This is about the Yarner-specific comments that will not end up in the source code at all. I just think that "extract the Yarner-specific comments into Yarner-specific mark-up in the target document files and strip it from the target source files" is too complicated and too fringe a feature to be useful especially as the same effect can be achieved by marking up the asides explicitly with exactly the same result in the generated human-readable document.

README.md
mlange-42: Or use string replace...

adamreichold: Yes I was anticipating to use have a placeholder in the template like `<<<MAIN_FILE>>>` that can then be processed using `template.replace("<<<MAIN_FILE>>>", actual_main_file)` at runtime?

adamreichold: It is admittedly a kluge, but it usually works out fine and with the templates you do know if they should every contain `<<<MAIN_FILE>>>` with a different meaning than the placeholder.

docs/Features.md
mlange-42: Just seen that there is already `default_language` in the toml.

README.md
mlange-42: Resolved...

src/bin/main.rs
```diff
@@ -355,34 +334,31 @@ fn create_project(file: &str, style: &str) -> Result<(), Box<dyn Error>> {
...
-            templates::BIRD
+            //config.html = Some(HtmlParser::default());
+            (templates::HTML, templates::HTML_CONFIG)
         }
-        _ => "",
+        _ => ("", ""),
     };
 
-    let toml = toml::to_string(&config).unwrap();
+    //let toml = toml::to_string(&config).unwrap();
```
adamreichold, line 360: Not a fan of commenting out code: If it is gone, it should be gone. It can always be retrieved using version control, in the current revision it is just noise.

PR #26, opened 2020-11-30
Release/0.1.3

mlange-42:
### Features

*  Prevent multiple writes to the same code file from different source files

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -170,8 +170,8 @@ pub trait Parser: ParserConfig {
         let pref = self.file_prefix();
         for (name, _block) in doc.tree().code_blocks(language) {
             if let Some(name) = name {
-                if name.starts_with(pref) {
-                    entries.push((name.to_owned(), (&name[pref.len()..]).to_owned()))
+                if let Some(rest) = name.strip_prefix(pref) {
+                    entries.push((name.to_owned(), rest.to_owned()))
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -186,8 +188,32 @@ pub enum ParseError {
     UnclosedTransclusionError(String),
     /// Error for invalid transclusions, e.g. if the file is not found
     InvalidTransclusionError(String),
+    /// Error for multiple locations with entrypoints to the same code file
+    MultipleCodeFileAccessError(String),
 } // is there even such a thing as a parse error? who knows.
 
+impl ParseError {
```
adamreichold, line 195: Just as a generic hint: It does not look like the variants of `ParseError` are ever matched since this is a CLI program that will probably always bubble up errors to eventually print the error messages, something like `Box<dyn Error>` might actually be a completely sufficient error type (both `&'static str` as well as `String` implement `Error`) to pass along custom error messages and errors from libraries. (For example, the [Jaccard agreement](https://git.ufz.de/ecoepi/jaccard-agreement/-/blob/main/src/main.rs) CLI uses this approach.)

mlange-42: @adamreichold You mean the differentiation between the different error causes is not necessary, since the user will read the message, and no code is ever using the error type?

adamreichold: Exactly. There might be a minuscule performance hit from boxing up the error,  but this could even be an improvement as the box is a single pointer in contrast with four pointers for the enum containing a `String`.

PR #28, opened 2020-12-14
Modify code paths

### LINKED ISSUE #27: [Feature] Allow different relative paths for copying "code" files.
Currently, code files to copy (toml option `code_files`) end in the same relative paths as where they were copied from. There should be an option to modify paths.

E.g., some files may need to end in the top level of code output, but should be in some sub-folder in the sources.

In progress under #28.

mlange-42:
@adamreichold I think this is a sufficiently simple solution which solves the cases encountered so far.

You probably have some improvement suggestions regarding borrowing, `to_owned` etc. Also, there may be a more idiomatic solution than the for loop in `modify_path`.

Fixes #27.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -186,42 +186,6 @@ fn main() {
         }
     };
 
-    match (&paths.code_files, &paths.code_paths) {
-        (Some(code_files), Some(code_paths)) if code_files.len() != code_paths.len() => {
-            eprintln!(
-            "If argument code_paths is given in the toml file, it must have as many elements as argument code_files",
-        );
-            return;
-        }
-        _ => (),
-    }
-
-    let code_files = match (&paths.code_files, &paths.code_paths) {
-        (Some(code_files), Some(code_paths)) => code_files
-            .iter()
-            .zip(code_paths)
-            .flat_map(|(file, replacement)| {
-                PathUtil::list_files(file)
-                    .unwrap()
-                    .into_iter()
-                    .map(move |path| (path.clone(), Some(modify_path(&path, &replacement[..]))))
-            })
-            .collect::<Vec<_>>(),
-        (Some(code_files), None) => PathUtil::list_all_files(&code_files[..])
-            .unwrap()
-            .into_iter()
-            .map(|path| (path, None))
-            .collect::<Vec<_>>(),
-        (None, _) => Vec::new(),
-    };
-
-    let doc_files = paths.doc_files.as_ref().map(|files| {
-        PathUtil::list_all_files(&files[..])
-            .unwrap()
-            .into_iter()
-            .collect::<Vec<_>>()
-    });
-
     for input in inputs {
         let (file_name, style_type, code_type) = {
             let file_name = PathBuf::from(&input);
@@ -334,47 +298,60 @@ fn main() {
         };
     }
 
+    match (&paths.code_files, &paths.code_paths) {
+        (Some(code_files), Some(code_paths)) if code_files.len() != code_paths.len() => {
+            eprintln!(
+                "If argument code_paths is given in the toml file, it must have as many elements as argument code_files",
+            );
+            return;
+        }
+        _ => (),
+    }
+
     let mut track_copy_dest = HashMap::new();
     if let Some(code_dir) = code_dir {
-        for (code_file, code_out_path) in &code_files {
-            let out_path = if let Some(out_path) = code_out_path {
-                match track_copy_dest.entry(out_path) {
-                    Occupied(entry) => {
-                        eprintln!(
-                            "ERROR: Attempted to copy multiple code files to {:?}: from {:?} and {:?}",
-                            out_path, entry.get(), code_file
-                        );
-                        return;
+        if let Some(code_file_patterns) = paths.code_files {
+            for (idx, code_file_pattern) in code_file_patterns.iter().enumerate() {
+                let code_path = paths.code_paths.as_ref().map(|code_paths| &code_paths[idx]);
+                for code_file in PathUtil::list_files(code_file_pattern).unwrap() {
+                    let out_path = code_path.map_or(code_file.clone(), |code_path| {
+                        modify_path(&code_file, &code_path)
+                    });
+                    match track_copy_dest.entry(out_path.clone()) {
+                        Occupied(entry) => {
+                            eprintln!(
+                                "ERROR: Attempted to copy multiple code files to {:?}: from {:?} and {:?}",
+                                out_path, entry.get(), code_file
+                            );
+                            return;
+                        }
+                        Vacant(entry) => {
+                            entry.insert(code_file.clone());
+                        }
                     }
-                    Vacant(entry) => {
-                        entry.insert(code_file);
-                        out_path
+                    let mut file_path = code_dir.clone();
+                    file_path.push(out_path.clone());
+
+                    eprintln!("Copying code file {:?} to {:?}", code_file, out_path);
+                    fs::create_dir_all(file_path.parent().unwrap()).unwrap();
+                    match fs::copy(&code_file, &file_path) {
+                        Ok(_) => {}
+                        Err(err) => eprintln!(
+                            "ERROR: --> Error copying code file {:?}: {}",
+                            code_file, err
+                        ),
                     }
                 }
-            } else {
-                code_file
-            };
-            let mut file_path = code_dir.clone();
-            file_path.push(out_path);
-
-            eprintln!("Copying code file {:?} to {:?}", code_file, out_path);
-            fs::create_dir_all(file_path.parent().unwrap()).unwrap();
-            match fs::copy(&code_file, &file_path) {
-                Ok(_) => {}
-                Err(err) => eprintln!(
-                    "ERROR: --> Error copying code file {:?}: {}",
-                    code_file, err
-                ),
             }
         }
     }
 
     if let Some(doc_dir) = doc_dir {
-        if let Some(doc_files) = doc_files {
-            for doc_file in &doc_files {
+        if let Some(doc_files) = paths.doc_files {
+            for doc_file in PathUtil::list_all_files(&doc_files[..]).unwrap() {
                 eprintln!("Copying doc file {:?}", doc_file);
                 let mut file_path = doc_dir.clone();
-                file_path.push(doc_file);
+                file_path.push(&doc_file);
                 fs::create_dir_all(file_path.parent().unwrap()).unwrap();
                 match fs::copy(&doc_file, &file_path) {
                     Ok(_) => {}
@@ -399,7 +376,7 @@ fn modify_path(path: &PathBuf, replace: &str) -> PathBuf {
         .collect();
     for (i, comp) in path.components().enumerate() {
         if repl_parts.len() > i && repl_parts[i] != "_" {
-            if repl_parts[i] != "." {
+            if repl_parts[i] != "-" {
                 new_path.push(repl_parts[i]);
             }
         } else {
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -233,7 +233,7 @@ fn main() {
                 ) {
                     eprintln!(
                         "ERROR: Failed to compile source file \"{}\": {}",
-                        file_name.to_str().unwrap(),
+                        file_name.display(),
                         error
                     );
                     continue;
@@ -259,7 +259,7 @@ fn main() {
                 ) {
                     eprintln!(
                         "ERROR: Failed to compile source file \"{}\": {}",
-                        file_name.to_str().unwrap(),
+                        file_name.display(),
                         error
                     );
                     continue;
@@ -285,7 +285,7 @@ fn main() {
                 ) {
                     eprintln!(
                         "ERROR: Failed to compile source file \"{}\": {}",
-                        file_name.to_str().unwrap(),
+                        file_name.display(),
                         error
                     );
                     continue;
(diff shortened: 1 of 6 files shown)


src/bin/main.rs
```diff
@@ -179,10 +185,31 @@ fn main() {
         }
     };
 
-    let code_files = paths.code_files.as_ref().map(|files| {
-        PathUtil::list_all_files(&files[..])
-            .unwrap()
-            .into_iter()
+    if let Some(code_files) = &paths.code_files {
```
adamreichold, line 188: It might be too dense, but you should be able to do this as a single `match`, e.g.
```rust
match (&paths.code_files, &paths.code_paths) {
  (Some(code_files), Some(code_paths)) if code_files.len() != code_paths.len() => { ... return; }
  _ => (),
}
```

src/bin/main.rs
```diff
@@ -179,10 +185,31 @@ fn main() {
...
+        if let Some(code_paths) = &paths.code_paths {
+            if code_files.len() != code_paths.len() {
+                eprintln!(
+                    "If argument code_paths is given in the toml file, it must have as many elements as argument code_files",
+                );
+                return;
+            }
+        }
+    }
+    let code_files = paths.code_files.as_ref().map(|patterns| {
```
adamreichold, line 198: I would suggest sticking to `code_files` for the closure argument as the newly introduced `patterns` made me stumble when reading this.

src/bin/main.rs
```diff
@@ -179,10 +185,31 @@ fn main() {
...
+        patterns
+            .iter()
+            .zip(
+                paths
+                    .code_paths
+                    .as_ref()
+                    .unwrap_or(&vec!["".to_string(); patterns.len()]),
```
adamreichold, line 205: I wonder if `zip` is really helping here, especially as it implies to create a fallback vector even though the lengths have been checked. Maybe using `enumerate` and indexing directly into `code_paths` (using `[..]` as it cannot panic due to the length check if it is at least `Some(..)`) is simpler?

src/bin/main.rs
adamreichold: It might also make sense to merge this preprocessing with the above length check code to reduce the number of times these `Option` values are unpacked, something like
```rust
let code_files = match (&path.code_files, &path.code_paths) {
  (Some(code_files), Some(code_paths)) => ...,
  (Some(code_files), None) => ...,
  (None, _) => Vec::new(),
};
```

src/templates/Yarner-tex.toml
```diff
@@ -37,6 +37,11 @@ docs = "docs/"
 files = ["%%MAIN_FILE%%"]
 ## File(s) to include in code output (unprocessed), as glob pattern(s).
 # code_files = ["**/*.rs"]
+## Allows to modify paths of copied code files. Replaces path components. Optional.
+## When given, must have as many elements as code_files.
+## E.g., use "foo/bar" to replace the first two path component by foo and bar.
```
adamreichold, line 42: Typo: "first two path components"

Concerning the design, I am somewhat skeptical about creating a dedicated mini language for this. If you go for the find-and-replace approach, regular expressions could do all of this and more while the reducing the cognitive overhead as most developers have used them in form or another.

src/bin/main.rs
```diff
@@ -328,27 +370,45 @@ fn main() {
...
+                    ),
                 }
             }
         }
     }
 }
 
+fn modify_path(path: &PathBuf, replace: &str) -> PathBuf {
+    let mut new_path = PathBuf::new();
+    let repl_parts: Vec<_> = replace.split('/').filter(|&x| !x.is_empty()).collect();
```
adamreichold, line 385: Did you try create a `Path` instance using `Path::new` and iterate its components instead of splitting manually here?

adamreichold: I think you also do not need to collect here, e.g.
```rust
let mut path = path.components();
let mut replace = Path::new(replace);
for (path, replace) in path.zip(replace) { ... }
for path in path { new_path.push(push); }
```

adamreichold: I also wonder whether a fast path for returning `path` unmodified would be a sensible addition?

src/bin/main.rs
```diff
@@ -305,20 +332,35 @@ fn main() {
         };
     }
 
+    let mut track_copy_dest = HashMap::new();
     if let Some(code_dir) = code_dir {
         if let Some(code_files) = code_files {
-            for code_file in &code_files {
-                eprintln!("Copying code file {:?}", code_file);
+            for (code_file, code_out_path) in &code_files {
+                if track_copy_dest.contains_key(code_out_path) {
```
adamreichold, line 339: Using `match track_copy_dest.entry(code_out_path)` could probably remove one of the hash map look-ups.

src/bin/main.rs
```diff
@@ -305,20 +332,35 @@ fn main() {
...
+                    eprintln!(
+                        "ERROR: Attempted to copy multiple code files to {:?}: from {:?} and {:?}",
+                        code_out_path, track_copy_dest[code_out_path], code_file
+                    );
+                    return;
+                }
+                track_copy_dest.insert(code_out_path, code_file);
+
                 let mut file_path = code_dir.clone();
```
adamreichold, line 348: Concerning efficiency, I wonder if it would make sense to delay resolving the glob patterns and doing the path replacement until here as the glob patterns could make the expanded list very large with the target paths adding memory overhead.

src/templates/Yarner-tex.toml
mlange-42: Personally, I am always struggling with RegEx and never really got it.

I would not see this as a mini-language and the only "special" here is the asterisk (which I may drop). Regarding the dot, the related part in the code is only for better readability of code output, the result would be the same (i.e. "." as "here"). Otherwise, it is just consecutive replacement of things between slashes.

src/bin/main.rs
mlange-42: The problem here is that the paths may be different. So, using the glob "**/*.txt" may result in a different first component.

src/bin/main.rs
mlange-42: Resolved through using your suggestion (at least the fallback vector problem)

```rust
let code_files = match (&path.code_files, &path.code_paths) {
  (Some(code_files), Some(code_paths)) => ...,
  (Some(code_files), None) => ...,
  (None, _) => Vec::new(),
};
```

src/bin/main.rs
mlange-42: Fast returning is good.

Regarding `Path` instead of splitting, it would probably not work due to dots being lost? I think I will have to try...

mlange-42: Don't get it working using `Path`.

src/bin/main.rs
adamreichold: But the expanded code files is essentially a `flat_map` over the code files. I am merely proposing to move this inner loop here without changing the logic at all (instead of collecting into a vector and having a single loop here).

src/bin/main.rs
adamreichold: Because dot has a special meaning for paths. I would probably suggest avoiding dot and asterisk then due to the ambiguity with the current directory and glob patterns. Maybe two other sigils could perform the same task?

adamreichold: (Nit `return path.clone();` would probably better convey the intent compared to `return path.to_owend();` as `PathBuf` is already owning. But of course `impl ToOwned for T where T: Clone`...)

mlange-42: > Because dot has a special meaning for paths.

That's what I mean with "not special", as it keeps it the normal meaning in paths. 😄 
I.e. if you leave out the dot part from the code, the result would be the same (except for `eprintln!`). Replacing the path component by a dot results in omitting the component.

mlange-42: The advantage I see for '*' is, that it is not a legal characher in paths. E.g. '_' would collide.

src/bin/main.rs
mlange-42: Ok, so I probably don't understand what you mean...

There is one replacement "pattern" per glob pattern. I iterate over globs and replacements "in parallel", so I always get a (glob, replacement) pair. I then get all the files for the glob, and do the replacement (using the same replacement "pattern") on every of those files.

Could you try to explain your suggestion in such a verbal way?

src/bin/main.rs
adamreichold: > Replacing the path component by a dot results in omitting the component.

While I can follow how to arrive at this interpretation, I find it much too clever as it is applied here. I think the design rationale for something that even a heavy user of Yarner will interact with maybe twice in a blue moon should be as simple and straight forward as possible. If using something complex but well know like regular expression is deemed inappropriate, going for the simple version that just lists all targets and sources (three in the case we are interested in) seems preferable IMHO.

But then again, we already presented these positions and hence let's leave it as it is.

> Personally, I see no problem in splitting "manually", so what is your reason to prefer the Path?

Using the existing primitive to interpret text as a path relays intent and makes sure to handle any special case, like platform-specific separators instead of slash.

> The advantage I see for '*' is, that it is not a legal characher in paths.

At least for POSIX, everything but / and \0 are legal characters in paths. Even more restricted versions generally restrict to valid UTF-8 or ASCII which still allows *. It is a glob pattern and has special meaning when interacting with a shell, but that does not prevent from being part of a path, i.e. "*" is also a valid file and directory name.

src/bin/main.rs
adamreichold: > Could you try to explain your suggestion in such a verbal way?

That part does not change at all. I am just suggesting to avoid collecting all the source-destination pairs in a vector before iterating over the entries of this vector to copy source to destination.

Instead of
```rust
for (code_file, code_out_path) in &code_files { /* copy code_file to code_out_path */ }
```
you could do something like
```rust
for (idx, code_file) in paths.code_files.iter().enumerate() {
  let code_path = paths.code_paths.as_ref().map(|code_paths| code_paths[idx]);

  for code_file in PathUtil::list_files(code_file) {
    let out_path = code_path.map_or(code_file, |code_path| modify_path(code_file, code_path));

    /* copy code_file to out_path */
  }
}
```

mlange-42: Ah, thanks!

src/bin/main.rs
mlange-42: For `Path`, dot etc. are not a problem it seems. It's just that I am not familiar with the `Components`, e.g. how to get their count, but I will just have to dig into the documentation deeper.

adamreichold: > e.g. how to get their count, but I will just have to dig into the documentation deeper.

While not efficient, `Components` is an iterator and hence `Iterator::count` can be used. But if you `.collect()` into `Vec<Component>`, this should not be an issue as you can just use `.len()`.

adamreichold: (The two-loops-zip version should also not need the length, should it?)

mlange-42: Think I got it with `Path`. So I will go for underscore instead of asterisk, although it would work too.

mlange-42: Ok, you were right. is works with a single dot, but not with more than one.

src/bin/main.rs
mlange-42: Implemented the suggestion, also for doc files. However, it was quite some fight against the borrow checker, and I probably have one or more "unnecessary" `clone`s there.

src/bin/main.rs
```diff
@@ -305,50 +298,108 @@ fn main() {
...
+                                out_path, entry.get(), code_file
+                            );
+                            return;
+                        }
+                        Vacant(entry) => {
+                            entry.insert(code_file.clone());
+                        }
+                    }
+                    let mut file_path = code_dir.clone();
+                    file_path.push(out_path.clone());
```
adamreichold, line 333: I think you can avoid this `.clone()` call if you move the `eprintln!` call above the construction of `file_path`.

src/bin/main.rs
```diff
@@ -305,50 +298,108 @@ fn main() {
...
+
+                    eprintln!("Copying code file {:?} to {:?}", code_file, out_path);
```
adamreichold, line 335: I just noticed that you either use `.to_str().unwrap()` or `{:?}` everywhere to print `Path`s. There is [`Path::display`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.display) specifically for that purpose.

src/bin/main.rs
```diff
@@ -305,50 +298,108 @@ fn main() {
...
                 }
             }
         }
     }
+
     if let Some(doc_dir) = doc_dir {
-        if let Some(doc_files) = doc_files {
-            for doc_file in &doc_files {
+        if let Some(doc_files) = paths.doc_files {
+            for doc_file in PathUtil::list_all_files(&doc_files[..]).unwrap() {
```
adamreichold, line 351: I wonder if you can avoid the slicing coercion by starting with
```rust
if let Some(doc_files) = &paths.doc_files {
```

src/bin/main.rs
adamreichold: You could still avoid collecting the components into a vector using the variant
```rust
let mut path = path.components();
let mut replace = Path::new(replace);
for (path, replace) in path.zip(replace.components().map(|c| c.as_os_str())) { ... }
for path in path { new_path.push(push); }
```
which also avoids the index checks.

src/bin/main.rs
```diff
@@ -305,51 +298,116 @@ fn main() {
...
     if let Some(doc_dir) = doc_dir {
-        if let Some(doc_files) = doc_files {
-            for doc_file in &doc_files {
-                eprintln!("Copying doc file {:?}", doc_file);
+        if let Some(doc_files) = &paths.doc_files {
+            for doc_file in PathUtil::list_all_files(doc_files).unwrap() {
+                eprintln!("Copying doc file {}", doc_file.display());
                 let mut file_path = doc_dir.clone();
-                file_path.push(doc_file);
+                file_path.push(&doc_file);
```
adamreichold, line 360: This seems to be taking a ref of a ref now?

mlange-42: No, doc_file is `PathBuf`. `PathUtil::list_all_files(doc_files)` returns `Result<Vec<PathBuf>`.

PR #31, opened 2020-12-17
Set root directory, doc file modification patterns

mlange-42:
Option to modify doc file copy path equivalently to code files.

Option to set the "root directory". Allows to have the "sources" in a sub-directory of the folder where `Yarner.toml` sits and where `yarner` is executed. Simpy changes the working directory.


src/bin/main.rs
```diff
@@ -178,14 +198,20 @@ fn main() {
...
                  > yarner -h",
-            );
-            return;
+            ));
         }
     };
 
+    if inputs.is_empty() {
+        return Err(format!(
+            "ERROR: No input files found. For help, use:\n\
```
mlange-42, line 210: @adamreichold Any idea why this does not produce a newline, neither in the console, nor when redirecting to a file?

src/bin/main.rs
```diff
@@ -12,6 +12,13 @@ use yarner::util::PathUtil;
 use yarner::{templates, MultipleTransclusionError, ProjectCreationError};
 
 fn main() {
+    match run() {
+        Ok(_) => {}
+        Err(err) => eprintln!("{}", err),
+    }
+}
```
mlange-42, line 19: @adamreichold Any better idea how to solve the problem that escape sequences in error messages are ignored when just returning `Result<...>` from main? If I don't "wrap" it into a function that prints the error, I get things like \n or \" instead of newline or quotes.

src/bin/main.rs
```diff
@@ -12,6 +12,12 @@ use yarner::util::PathUtil;
 use yarner::{templates, MultipleTransclusionError, ProjectCreationError};
 
 fn main() {
+    if let Err(err) = run() {
+        eprintln!("{}", err);
+    }
```
mlange-42, line 17: @adamreichold Any better idea how to solve the problem that escape sequences in error messages are ignored when just returning `Result<...>` from main? If I don't "wrap" it into a function that prints the error, I get things like \n or \" instead of newline or quotes.

adamreichold: You could create a custom error type and do any formatting in its `Debug` implementation to avoid this wrapping, but I would consider this a hack. If you do want control over how the errors are printed, I would say that this is the direct and also idiomatic way.  But note that you should still adjust you exit code here, i.e. use `fn main() -> i32` as the program should not return zero on failure. Otherwise, (CI) scripts will not be able to react to these errors.

(On a more general note, I think Yarner could have a much simpler code base if we would reconsider some of the design choices inherited from Outline, e.g. use simple boxed errors, drop the library/binary split, etc. I guess an overhaul or even a rewrite would be nice holiday project?)

src/bin/main.rs
```diff
@@ -36,6 +42,12 @@ fn main() {
             .help("Sets the style to use. If not specified, it is inferred from the file extension.")
             .takes_value(true)
             .possible_values(&["md", "tex", "html"]))
+        .arg(Arg::with_name("root")
+            .long("root")
+            .short("r")
+            .value_name("root")
+            .help("Root directory. If none is specified, uses 'path' -> 'root' from config file. Default current director.")
```
adamreichold, line 49: Typo: "director" instead of "directory"

src/bin/main.rs
```diff
@@ -114,27 +126,38 @@ fn main() {
...
-
     let paths = any_config.paths.unwrap_or_default();
 
+    let root = matches
+        .value_of("root")
+        .map(|s| s.to_string())
+        .or_else(|| paths.root.as_ref().map(|s| s.to_string()))
+        .map_or_else(|| PathBuf::from("."), |r| PathBuf::from(r));
+
+    if let Err(err) = std::env::set_current_dir(&root) {
```
adamreichold, line 153: I am not so sure this option is pulling its weight if this is its only effect. One can easily change the working directory in the script that calls Yarner. But I guess you arrived at this with a particular use case in mind?

src/bin/main.rs
```diff
@@ -298,78 +327,74 @@ fn main() {
...
-                    match fs::copy(&code_file, &file_path) {
-                        Ok(_) => {}
-                        Err(err) => eprintln!(
-                            "ERROR: --> Error copying code file {}: {}",
-                            code_file.display(),
-                            err
-                        ),
-                    }
+    for (idx, file_pattern) in patterns.iter().enumerate() {
+        let code_path = path_mod.as_ref().map(|code_paths| &code_paths[idx]);
```
adamreichold, line 360: I think it would improve readability to excise all `code_*` prefix from this now general code?

src/bin/main.rs
```diff
@@ -298,78 +327,74 @@ fn main() {
...
+        for code_file in PathUtil::list_files(file_pattern).unwrap() {
```
adamreichold, line 361: If you are, cleaning this up: I don't `PathUtil` is actually a useful thing. For example in this case, it forces collecting the glob matching results into vector and it swallows any errors encountered during globbing. I think just using `glob::glob` directly here would be simpler and more efficient. (I think ideally `PathUtil` would go away completely, but this sort of leads to the whole overhaul/rewrite discussion again...)

src/bin/main.rs
mlange-42: As returning `i32` is not possible, I would use this:

```rust
fn main() {
    std::process::exit(match run() {
        Ok(_) => 0,
        Err(err) => {
            eprintln!("{}", err);
            1
        }
    });
}
```

mlange-42: I also think that a general review and rework of the code would make sense, although I would probably not do a complete rewrite. Some things can probably also be removed like the html (definitely) and tex (maybe) parsers.

src/bin/main.rs
mlange-42: I stumbled when trying to set up a small test project combining Yarner and MdBook. Wanting Yarner sources in a sub-directory would require the `Yarner.toml` to be in the sub-directory too (in contrast to `book.toml`), and multiple `cd`-ing to navigate around. Otherwise, doc output would end in a sub-sub-folder (i.e. a sub-folder of `docs`).

src/bin/main.rs
mlange-42: The error handling is indeed suboptimal here, and could definitely be improved. I will first clean it from unused functions, and maybe try to get rid of it.

src/bin/main.rs
adamreichold: But then maybe it would be nicer to support having Yarner.toml somewhere else besides the working directory (like Cargo's manifest-path parameter)?

src/bin/main.rs
```diff
@@ -159,34 +185,56 @@ fn main() {
...
-            );
-            return;
+            ))
         }
+        Some(patterns) => patterns,
     };
 
-    for input in inputs {
+    let mut inputs: Vec<PathBuf> = Vec::new();
+    for pattern in input_patterns {
```
mlange-42, line 204: Due to the "deep boxing" of results in vectors in results in option (or similar) I was not able to solve this using comprehensions, so it does not look that "rusty". However, I got rid of `PathUtil`.

src/bin/main.rs
mlange-42: Got rid of `PathUtil` and probably handle errors appropriately now. But see https://github.com/mlange-42/yarner/pull/31#discussion_r545816458.

src/bin/main.rs
mlange-42: But for the structure I intended that would mean to navigate into the sub-directory, and then "call" the toml file from a level above. Hm, this is even more typing and fairly confusing...

adamreichold: At least with Cargo, the idea is that all paths in the manifest at relative to the path of the manifest. The manifest-path argument then allows one to start Cargo from anywhere by pointing it at a manifest.

adamreichold: > Otherwise, doc output would end in a sub-sub-folder (i.e. a sub-folder of docs).

Why can't the documentation output path include a ".." to end up higher in the folder hierarchy?

mlange-42: It can, but that does not solve the problem. Here is an example what happens with the current (i.e. old) implementation. We start with this structure:

```
project
  |--  src/
  |      |--  File-1.md
  |      `--  File-2.md
  `--  Yarner.toml
```

Let's say output folders are `code` and `docs`. We get this:

```
project
  |--  code/
  |--  docs
  |      `--  src
  |             |--  File-1.md
  |             `--  File-2.md
  |--  src/
  |      |--  File-1.md
  |      `--  File-1.md
  `--  Yarner.toml
```

What works is this, with `Yarner.toml` in `src`, running the command from `src` and having output paths `../code` and `../docs`:

```
project
  `--  src/
         |--  File-1.md
         |--  File-1.md
         `--  Yarner.toml
```

Results in this structure:

```
project
  |--  code/
  |--  docs
  |       |--  File-1.md
  |       `--  File-2.md
  `--  src/
         |--  File-1.md
         |--  File-1.md
         `--  Yarner.toml
```

adamreichold: > IMO more severe, "hides" the Yarner.toml.

Yes, this is not helpful. I also see now that there is difference to changing the working directory via the shell: The point here is to change it after loading `Yarner.toml`. So all paths are relative to root, except for `Yarner.toml` which is in the folder to which `root` is relative to.

You could consider using the same mechanism as you already employ for `code_files` and `doc_files` also for `files` to e.g. strip the first path component to the keep the number of distinct mechanisms smaller, gain a bit of flexibility and increase the return on investment for learning the path manipulation language.

mlange-42: Yes I was already thinking about that option. First, I liked the option for the higher flexibility, but one could also break hyperlinks between doc files when defining different replacement "patterns" for different files/globs. I was not really happy with that, but maybe I should think about it again...

src/bin/main.rs
mlange-42: Moved file procesing into that loop, so no storage for inputs in vector anymore.

src/bin/main.rs
mlange-42: As path modifications like for copying code and doc files would also impact (or conflict with) transclusions and link-following, I with stay with the current solution.

PR #33, opened 2021-01-15
fix cargo install command in README, fix compatibility break of serde

mlange-42:
* Fix cargo install command in README
* Fix compatibility break of serde (`serde::export` no more public)

### DIFF
--- a/src/document/text.rs
+++ b/src/document/text.rs
@@ -1,6 +1,6 @@
 //! Representation of the text parts of the AST
 
-use serde::export::Formatter;
+use core::fmt::Formatter;
 use std::fmt::Display;
 
 /// A `TextBlock` is just text that will be copied verbatim into the output documentation file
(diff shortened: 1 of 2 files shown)


src/document/text.rs
```diff
@@ -1,6 +1,6 @@
 //! Representation of the text parts of the AST
 
-use serde::export::Formatter;
+use core::fmt::Formatter;
```
adamreichold, line 3: Is seems a bit confusing to import parts of `fmt` from `core` others from `std`. Better to stick to using only `std::fmt`. (`core::fmt` is only useful if you want to build a `no_std` crate.)

mlange-42: Ok. Read it here, but it was a misinterpretation of the slash between `std` and `core`: https://users.rust-lang.org/t/serde-compatibility-broken/54041/2

Would it be a problem to stay with `core` until the next release?

adamreichold: > Would it be a problem to stay with core until the next release?

No, that is not an issue. Everything in `std::fmt` that is also in `core::fmt` is actually a re-export of the `core` item in the `std` module. (But the `std` module contains additional items that require an allocator or operating system.)

mlange-42: Ok, thanks. So I'll leave the change for the next release.

PR #41, opened 2021-01-25
Yarner documentation as mdBook

### LINKED ISSUE #21: [Documentation] Write a User Guide. Full docs, incl. side topics.
**[User Guide](https://mlange-42.github.io/yarner/)**

Possible sections:

* [x] Yarner syntax and features
* [x] Configuration
* [x] CI/CD examples/templates (separate issue: #50)
* [x] Managing multi-file projects: recommendations for project structure under diffenent use cases
* [x] Markdown tool recommendations
  * [x] MarkText MD editor
  * [x] Atom
  * [x] Math formulas: RStudio as MD editor
  * [x] For GitHub, see [A hack for showing LaTeX formulas in GitHub markdown.md](https://gist.github.com/a-rodin/fef3f543412d6e1ec5b6cf55bf197d7b)
  * [x] Conversion tool: Pandoc (to HTML, PDF, DOCX, ...)
* [x] Usage examples
  * [x] mdBook with Yarner

mlange-42:
* [x] Setup empty book and CI
* [x] Outline structure
* [x] Write chapters

Fixes #21


guide/src/continuous-integration.md
```diff
@@ -0,0 +1,38 @@
+# Continuous integration
```
mlange-42, line 1: @adamreichold I would highly appreciate some help on this chapter to present some minimal examples for CI usage on these most common providers.

guide/src/SUMMARY.md
```diff
@@ -0,0 +1,23 @@
+# Summary
+
+[Yarner -- Introduction](./introduction.md)
+
+<!-- - [How is works](./how-it-works.md) -->
```
adamreichold, line 5: Left over comment?

mlange-42: The chapter is still there, but I am currently not sure if it is useful at all, or if the introductions explains it already. Also, If I include it, I am also not sure if it should be the first (numbered) chapter, or if it should be placed after "Getting started" in order to avoid hindrance for people to just jump into trying it out.

adamreichold: From skimming both, I would suggest folding it into the introduction.

guide/src/command-line-arguments.md
```diff
@@ -0,0 +1,49 @@
...
+  https://github.com/mlange-42/yarner
+
+The normal workflow is:
+ 1) Create a project with
+    > yarner create README.md
+ 2) Process the project by running
+    > yarner
+
+USAGE:
+    yarner.exe [FLAGS] [OPTIONS] [input]... [SUBCOMMAND]
```
adamreichold, line 22: Is that suffix really part of the usage? If so, I think removing it would be preferable both here and there: Windows works without it and it is misleading on other platforms.

guide/src/continuous-integration.md
```diff
@@ -0,0 +1,38 @@
...
+
+This chapter shows some examples how to use Yarner with Continuous integration.
+There are always two ways to get Yarner:
+
+1. Use a rust image and install it using `cargo`
```
adamreichold, line 6: I would not recommend this at all. For a user of Yarner, it should not matter at all in which language it is implemented and if his/her project does not use Rust, he should not include this into its CI due to the diligence required to keep a CI rolling at all.

guide/src/continuous-integration.md
```diff
@@ -0,0 +1,38 @@
...
+## GitLab CI
+
+In your GitLab project, create a file `.gitlab-ci-yml` with the following content:
+
+```yml
+variables:
+  YARNER_VERSION: 0.2.2
+
+before_script:
+  - curl -L -o- https://github.com/mlange-42/yarner/releases/download/${YARNER_VERSION}/yarner-${YARNER_VERSION}-linux.tar.gz | tar xz
```
adamreichold, line 26: This is not so much CI-specific but a general question. Does the archive contain anything but the Yarner binary? If not, maybe we should just drop the subfolder within the archive?

guide/src/continuous-integration.md
```diff
@@ -0,0 +1,38 @@
...
+  - export PATH="$PWD/yarner:$PATH"
```
adamreichold, line 27: We could also move the Yarner binary to `/usr/local/bin` instead of adjusting the search path. This might be preferable to  including the project root in it because it might contain files colliding with other binary names.

guide/src/continuous-integration.md
adamreichold: Ok, but let's do this as follow-up PR. I think the GitLab CI example is basically sound and it is just a matter of porting it to GH Actions and Travis CI. I also do not think we should include more as these three are already quite redundant.

What would maybe be interesting is an example of how to publish the results, e.g. render the Markdown to HTML using Pandoc and push that to GH Pages?

mlange-42: Completely agree on all your points. I did not plan do include more than these three. An example for publishing is definitely a good idea.

guide/src/command-line-arguments.md
mlange-42: Ah, yes, it is there because I simply copied to path of `target/debug/yarner.exe` to paste it to the command line. Presence of the suffix depends on whether it is called with or without extension, and clap generates this line accordingly.

guide/src/continuous-integration.md
mlange-42: Removed it.

guide/src/continuous-integration.md
mlange-42: Yes, its only the executable. I did it this way because I don't like seeing files tumbling at me when extracting an archive, and normally expect things to be within a folder. Also, in the current state it is more convenient for simly moving the folder to tha programs folder. So I am not entirely convinced yet.

guide/src/continuous-integration.md
mlange-42: Sounds good, but I would suggest to leave that for the CI MR.

guide/src/SUMMARY.md
mlange-42: I decided to remove it. Most of the content is already present in `Getting started`.

guide/src/continuous-integration.md
adamreichold: Fair enough.

PR #44, opened 2021-01-29
Do some minor chores

adamreichold:
- [x] Update the locked Cargo dependencies
- [x] Fix all warnings thrown by today's nightly Clippy
- [x] Remove the lib/main split to reduce build duration
- [x] Switch from speed to size optimization as this application IO-bound rather than CPU-bound


adamreichold:
@mlange-42 A general recommendation is to run the CI against the latest nightly (but allowing build failures) to get a heads up on new lints or other upcoming changes. But I would probably do this only while also switching to GitHub Actions as well due to the increased build effort...

mlange-42:
@adamreichold Many thanks! I just renamed the branch to `hotfix/chores`, but that closed the pull request. You may want to create a new one linking here, or I can do so.

adamreichold:
I forgot this is doing the whole Git Flow rigamarole, I recreated this as `feature/chores` to be merged into `dev` then as it is not really fixing anything.

mlange-42:
😆 Great, thanks!

PR #45, opened 2021-01-29
Do some minor chores

adamreichold:
- [x] Update the locked Cargo dependencies
- [x] Fix all warnings thrown by today's nightly Clippy
- [x] Remove the lib/main split to reduce build duration
- [x] Switch from speed to size optimization as this application IO-bound rather than CPU-bound


mlange-42 (approved):

mlange-42:
@adamreichold You earned a round of beer 🍻

PR #46, opened 2021-01-29
Remove concrete abstractions

adamreichold:
I guess this more controversial as it not does not simplify things very much as it just removes traits that have only a single implementation and only starts on the way towards standardizing all error types to `Box<dyn Error>`. But then again, all the Markdown parsing code is now in place and simplification can start in earnest and it is a first step towards dropping the library-like heritage of much of the code which IMHO is unhelpful over-generalization for the program by now.


mlange-42:
I was also thinking about dropping all the now unnecessary traits in `parser/mod.rs`. As I don't expect other parsers than Markdown to be added soon (or at all), I am happy with that.

mlange-42 (approved):

PR #47, opened 2021-01-29
Simplify the util module

adamreichold:
This variant of `try_collect` might not be as general as the previous one, but it also avoids the `into_iter().collect()` roundtrip and reallocation for the `C_ = Vec<_>` case, which is also the only case used in the application.


mlange-42 (approved):

mlange-42:
Maybe you merge these, as you may be better able to resolve conflicts or do a rebase? Do you have the required permissions?

adamreichold:
> Maybe you merge these, as you may be better able to resolve conflicts or do a rebase? Do you have the required permissions?

I can do and will do. Generally speaking, the maintainer usually merges in the order he prefers. If this creates conflicts in the remaining pull requests, it is the task of their authors to resolve those. Nothing you have to do in that case besides possibly reminding people to go resolve those conflicts in their code.

PR #48, opened 2021-01-29
Start simplifing the Markdown parser implementation.

adamreichold:


src/parser/md.rs
```diff
@@ -149,53 +157,34 @@ impl Default for MdParser {
...
-        }
+        cloned
     }
 
     fn parse_transclusion(&self, line: &str, into: &Path) -> Fallible<Option<Node>> {
-        let trim = line.trim();
-        if trim.starts_with(&self.transclusion_start) {
-            if let Some(index) = line.find(&self.transclusion_end) {
-                let trans = &trim[self.transclusion_start.len()..index];
-                let regex = Regex::new(Self::LINK_PATTERN).unwrap();
```
mlange-42, line 184: Oh, I see that this was quite an evil line with a lot of unnecessary overhead.

mlange-42 (approved):

src/parser/md.rs
adamreichold: I guess it depends: The regular expression was only compiled when the transclusion prefix is found, so it is not done for every single line. But of course, generally speaking one should not try to compile regular expression as seldomly as possible which in this case of a constant pattern is once.

(I think this snippet did actually contain a bug though: `index` refers to the full `line` but is used to index into `trim`, i.e. if there is a lot of whitespace at the beginning and `line` is much longer than `trim`, this indexing could actually panic.)

PR #49, opened 2021-01-29
Replace more error types by Fallibe.

adamreichold:
`Fallible` might not look like a clear win ATM due to the additional `.into()` calls, but I think this will change when error propagation is used more systematically. For example,
```rust
match fs::read_to_string(file_name) {
    Ok(config) => match toml::from_str(&config) {
        Ok(config) => config,
        Err(error) => {
            return Err(format!(
                "ERROR: Could not parse config file \"{}\": {}",
                file_name, error
            )
            .into());
        }
    },
    Err(error) => {
        return Err(format!(
            "ERROR: Could not read config file \"{}\": {}",
            file_name, error
        )
        .into());
    }
}
```
could be done as
```rust
fn read_config(path: &Path) -> Fallible<AnyConfig> {
  let buf = fs::read_to_string(path)?;
  let val = toml::from_str(&buf)?;
  Ok(val)
}
...
let config = read_config(path).map_err(|err| format!("ERROR: Could not read config file \"{}\": {}", path.display(), err))?
```
which depends on usign `Box<dyn Error>` to unify `std::io::Error` and `toml::de::Error`, but I did not want to mix to many changes into one PR. (I am also not I understand what the `occurrences_of == 0` check above that snippet actually does.)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,15 +14,13 @@ use crate::util::Fallible;
 use clap::{crate_version, App, Arg, SubCommand};
 use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::{HashMap, HashSet};
-use std::error::Error;
-use std::fmt;
 use std::fs::{self, File};
 use std::io::Write;
 use std::path::{Path, PathBuf};
 
 fn main() {
     std::process::exit(match run() {
-        Ok(_) => 0,
+        Ok(()) => 0,
         Err(err) => {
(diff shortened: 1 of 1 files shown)


mlange-42 (approved):

mlange-42:
> I am also not I understand what the `occurrences_of == 0` check above that snippet actually does.

@adamreichold You are right, that case should be already covered by the `None` branch of the enclosing `match`.

PR #52, opened 2021-01-29
Link config chapters in README, fix left-over comment in chapter path configuration

mlange-42:


README.md
```diff
@@ -8,6 +8,16 @@ Yarner works with familiar syntax, which can be further customized to suit your
...
 
+## Features
+
+* Macros and named code blocks
+* Multiple files, multiple entrypoints
+* File transclusions
+* Reverse mode
+* ...
+
+See the [**User Guide**](https://mlange-42.github.io/yarner/) for a complete and detailed explanation of all features.
```
adamreichold, line 19: I wonder whether it would make sense to slim down the README now that the book exists? But admittedly from a first glance, I have no section which I would be comfortable with removing...

guide/src/installation.md
```diff
@@ -11,7 +11,7 @@ There are multiple ways to install Yarner:
 2. Unzip somewhere
 3. *Optional:* add directory `yarner` to your `PATH` environmental variable
 
-## From Git using `cargo`
+## From GitHub using `cargo`
```
adamreichold, line 14: I really think we should publish to crates.io to simplify things for potential users.

README.md
mlange-42: I already reduced it a bit. But I would still leave the basics here: installation, first steps/getting started, and an illustrative example. Maybe the entrypoints section, but I am not really sure about. 

Maybe we should let it read someone completely new to Yarner.

guide/src/installation.md
mlange-42: Now that you started to clean up the code, sure! 😉

adamreichold: So, this should go into a CI MR as you we should probably publish tags automatically via CI...

PR #54, opened 2021-01-30
Simplify config module

adamreichold:
A lot of changes but all are purely clean up without any behavior modifications, except for removing the ERROR and --> prefixes from the program output and standardizing on "ERROR" in the main handler.

Note that as a general recommendation, `Option<&Borrowed>` is preferable to `&Option<Onwed>` as a function parameter. One could always go from the latter to the former via `.as_deref()`, but the other way around requires a temporary allocation.

Also note that I did not try to avoid cloning `config.paths` so much because of the allocations, but rather to avoid having to places in the program state where the same information is stored as this allows one to introduce bugs by modifying one but reading from the other.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,7 +4,7 @@ pub mod parser;
 pub mod templates;
 pub mod util;
 
-use crate::config::{AnyConfig, LanguageSettings};
+use crate::config::{Config, LanguageSettings};
 use crate::document::{CompileError, CompileErrorKind, Document};
 use crate::parser::{
     code::{CodeParser, RevCodeBlock},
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -31,13 +33,13 @@ fn main() {
...
                   https://github.com/mlange-42/yarner\n\
                 \n\
-                The normal workflow is:\n \
-                1) Create a project with\n    \
-                  > yarner create README.md\n \
-                2) Process the project by running\n    \
+                The normal workflow is:\n\
+                1) Create a project with\n\
+                  > yarner create README.md\n\
+                2) Process the project by running\n\
```
mlange-42, line 42: Don't you think this was much better readable with indentations?

adamreichold: Ah, I was assuming the indentation would come from the next line as with doc comments, but that is not actually the case. I'll go for raw string literals then as removing the indentation was certainly not my intent.

adamreichold: Fixed using raw string literal.

mlange-42 (approved):

PR #55, opened 2021-01-30
Remove AST module

adamreichold:
Because it adds nothing but another mental indirection making it harder to follow what is going on.

Note that this depends #54 syntactically but not semantically, hence the chosen target branch instead of `dev`.

Also note that `&Option<&T>` is usually not helpful as `&T` and thereby `Option<&T>` are `Copy` and due to niche optimization (`&T` can never be null in Rust) have the same size, i.e. `&Option<&T>` adds nothing but another layer of indirection which can only worsen performance if it is not compiled away.


mlange-42 (approved):

mlange-42:
Thanks for the hints!

mlange-42 (approved):

src/parser/md.rs
```diff
@@ -557,12 +554,12 @@ impl MdParser {
...
         &self,
         doc: &Document,
-        language: &Option<&str>,
+        language: Option<&str>,
     ) -> Vec<(String, String)> {
         let mut entries = vec![];
         let pref = &self.file_prefix;
-        for (name, _block) in doc.tree().code_blocks(language) {
-            if let Some(name) = name {
+        for block in doc.code_blocks(language) {
```
mlange-42, line 561: @adamreichold A bug was introduced by this change, and by the change to the return type of function `code_blocks`. There was a reason to have the map... Currently investigating and trying to fix it.

PR #60, opened 2021-01-30
Reverted to use the (name, blocks) map to collect entrypoints

mlange-42:
Required to allow for multiple equal-named blocks that are entrypoints.

### DIFF
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -10,6 +10,7 @@ use self::transclusion::Transclusion;
 use crate::config::LanguageSettings;
 use crate::parser::{code::RevCodeBlock, md::MdParser};
 use std::collections::hash_map::{Entry, HashMap};
+use std::fmt::Write;
 
 /// A representation of a `Document` of literate code
 #[derive(Debug)]
@@ -161,25 +162,27 @@ impl Document {
                     };
 
                     if !clean {
-                        if idx == 0 || block.name != blocks[idx - 1].name {
-                            result.push_str(&format!(
-                                "{} {}{}#{}{}\n",
-                                comment_start, block_start, path, name, comment_end,
-                            ));
+                        let sep = if idx == 0 || block.name != blocks[idx - 1].name {
+                            &block_start
                         } else {
-                            result.push_str(&format!(
-                                "{} {}{}#{}{}\n",
-                                comment_start, block_next, path, name, comment_end,
-                            ));
-                        }
+                            &block_next
+                        };
+                        writeln!(
+                            result,
+                            "{} {}{}#{}{}",
+                            comment_start, sep, path, name, comment_end,
+                        )
+                        .unwrap();
                     }
                     result.push_str(&block.compile(&code_blocks, settings)?);
                     result.push('\n');
                     if !clean && (idx == blocks.len() - 1 || block.name != blocks[idx + 1].name) {
-                        result.push_str(&format!(
+                        write!(
+                            result,
                             "{} {}{}#{}{}",
                             comment_start, block_end, path, name, comment_end,
-                        ));
+                        )
+                        .unwrap();
                     }
                 }
             }
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -241,7 +241,7 @@ impl Document {
         })
     }
 
-    pub fn code_blocks_by_name<'a>(
+    fn code_blocks_by_name<'a>(
         &'a self,
         language: Option<&'a str>,
     ) -> HashMap<Option<&'a str>, Vec<&'a CodeBlock>> {
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -140,6 +140,7 @@ impl Document {
             .unwrap_or("");
         let block_start = settings.map(|s| &s.block_start[..]).unwrap_or("");
         let block_end = settings.map(|s| &s.block_end[..]).unwrap_or("");
+        let block_next = settings.map(|s| &s.block_next[..]).unwrap_or("");
 
         let clean = if let Some(s) = settings {
             s.clean_code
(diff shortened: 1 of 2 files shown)


src/parser/md.rs
```diff
@@ -558,8 +558,8 @@ impl MdParser {
     ) -> Vec<(String, String)> {
         let mut entries = vec![];
         let pref = &self.file_prefix;
-        for block in doc.code_blocks(language) {
-            if let Some(name) = &block.name {
+        for (name, _block) in doc.code_blocks_by_name(language) {
```
adamreichold, line 561: I think it would be a better change to make `entries` a `HashSet<(&str, &str)>` here instead of building the intermediate `HashMap` that is basically discarded.

src/document/mod.rs
```diff
@@ -151,7 +152,9 @@ impl Document {
         let mut result = String::new();
         match code_blocks.get(&entrypoint) {
             Some(blocks) => {
-                for block in blocks {
+                for idx in 0..blocks.len() {
```
adamreichold, line 155: `blocks.iter().enumerate()`?

src/parser/md.rs
mlange-42: I.e. add another function for only getting the names? Because the map is required in printing.

src/document/mod.rs
```diff
@@ -160,14 +163,21 @@ impl Document {
                     };
 
                     if !clean {
-                        result.push_str(&format!(
-                            "{} {}{}#{}{}\n",
-                            comment_start, block_start, path, name, comment_end,
-                        ));
+                        if idx == 0 || block.name != blocks[idx - 1].name {
+                            result.push_str(&format!(
```
adamreichold, line 167: Maybe just determine `let block_sep = if idx == 0 || ...` and then do
```rust
writeln!(&mut result, "{} {}{}#{}{}", comment_start, block_sep, path, name, comment_end).unwrap();
```
here? (`push_str` and `format!` means allocating a temporary string instead of writing directly into `result`.)

src/document/mod.rs
mlange-42: But don't I then get a problem to borrow indexed blocks? Probably not as long as I do no mutable borrows. I will try...

src/parser/md.rs
adamreichold: I would just change the return type of `get_entry_points` to `HashSet<(&str, &str)>` (or maybe the owned variant of that). (This cannot depend on order as `code_blocks_by_name` breaks order anyway.)

src/document/mod.rs
mlange-42: Is it safe to use `unwrap()` when writing to a string? Just asking because I didn't venture to do it in git-graph/git-igitt.

adamreichold: Definitely. The general `fmt::Write` interface needs to propagate errors when writing into files and such, but writing into a growable `String` can only fail due to out of memory and this aborts the process anyway.

src/parser/md.rs
mlange-42: Sure, but this means returning a HashSet instead of a Vector, but your original comment was about to avoid the HashMap. Which I thought referred to the return type of `code_blocks_by_name`, as there is no other HashMap involved.

adamreichold: > but your original comment was about to avoid the HashMap.

I meant to avoid the `HashMap` returned by `code_blocks_by_name` which is thrown away and instead using a `HashSet` that stays around replacing the `Vec` that stays around. `get_entry_points` does not need to call `code_blocks_by_name` at all just to collapse duplicate names, it can do that itself, e.g.
```rust
pub fn get_entry_points<'a>(
    &self,
    doc: &'a Document,
    language: Option<&'a str>,
) -> HashMap<&'a str, &'a str> {
    let mut entries = HashMap::new();
    let pref = &self.file_prefix;
    for block in doc.code_blocks(language) {
        if let Some(name) = &block.name {
            if let Some(rest) = name.strip_prefix(pref) {
                entries.insert(name.as_str(), rest);
            }
        }
    }
    entries
}
```

adamreichold: (This this does not even seem to require any changes at the call sites.)

mlange-42: Ah, now I got it. Ignore my the commit I just made ;)

mlange-42: It is not 100% equivalent, as the check for dublicates is now done after stripping the prefix. Not sure if this could be a problem. Only when there are equally-named code blocks, but one with `file:` and one without. In theory, that would be allowed.

adamreichold: > It is not 100% equivalent, as the check for dublicates is now done after stripping the prefix. Not sure if this could be a problem.

At least in the code above, the full name is used as the key so duplicates are checked exactly as before.

mlange-42: But on the other hand, it seems blocks without file prefix don't even arrive here. Not sure but I guess an unnamed entrypoint already erts renamed to `file:<target>` already before.

adamreichold: > Not sure but I guess an unnamed entrypoint gets renamed to file:<target> already before.

Isn't this
```rust
    let mut entries = vec![(entrypoint.as_deref(), file_name.to_owned())];
    let extra_entries = parser.get_entry_points(&document, language);

    entries.extend(
        extra_entries
            .iter()
            .map(|(e, p)| (Some(&e[..]), PathBuf::from((*p).to_owned() + ".temp"))),
    );
```
?

mlange-42: It does...

adamreichold: > Ah, now I got it. Ignore my the commit I just made ;)

This refers to the commit to collects into a `HashSet` and then into a `Vec`?

mlange-42: Yes. I think this is resolved not with my force-push.

adamreichold (approved):

PR #61, opened 2021-01-30
Fix space after macro call start, use indent string as found in code

mlange-42:
Minor fix of spaces and indentation in reverse mode

### DIFF
--- a/src/parser/code.rs
+++ b/src/parser/code.rs
@@ -89,9 +89,14 @@ impl CodeParser {
                 if !is_next {
                     if let (Some(name), Some(block)) = (&name, block_stack.last_mut()) {
                         let new_line = format!(
-                            "{}{} {}{}",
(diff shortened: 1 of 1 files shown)


adamreichold:
Clippy is not happy. This also contains commits from #60, so the target branch should probably change.

mlange-42:
Fixed clippy, changed target branch

adamreichold:
Needs a rebase...

mlange-42:
I can give it a try, but not sure if I will manage that.

adamreichold:
> I can give it a try, but not sure if I will manage that.

In this case, you'll have to as this contains an outdated commit that you force-pushed in #60. Just do `rebase -i origin/dev` and drop the first commit because it is not needed anymore.

adamreichold (approved):

PR #62, opened 2021-01-30
Get rid of default config, sort options in template by topic and importance

mlange-42:


src/parser/md.rs
```diff
@@ -137,10 +137,10 @@ impl Default for MdParser {
             fence_sequence: String::from("```"),
             fence_sequence_alt: String::from("~~~"),
             comments_as_aside: false,
-            comment_start: String::from("//"),
+            comment_start: String::from("//-"),
             interpolation_start: String::from("@{"),
             interpolation_end: String::from("}"),
-            macro_start: String::from("==> "),
+            macro_start: String::from("// ==>"),
```
adamreichold, line 143: Do want to keep the defaults in the config, why do we duplicate them in the code at all?

I am also a bit confused by the default config loading in the application. Is it actually possible to successfully build project without a config resp. with the default config? Shouldn't we just always load the (defaulting via Clap config) config file path and drop the defaults here?

mlange-42: Indeed, that would be possible. It is probably from before using the template, which was introduced to have comments in the generated config.

mlange-42: I guess I can `unwrap` this: `matches.value_of("config")` as a default value is given?

mlange-42: ... as in the current version...

adamreichold (approved):

PR #63, opened 2021-01-30
Avoid adding and then removing the .temp extension for entrypoint output paths

adamreichold:
@mlange-42 It is likely I am missing something, but it looks like we add `.temp` to all output paths allocating new `PathBuf`s along the way only to remove this again by calling `sub_file_name.file_stem().unwrap()`?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -724,14 +724,10 @@ fn compile(
 ) -> Fallible {
     eprintln!("Compiling file {}", file_name.display());
 
-    let mut entries = vec![(entrypoint.as_deref(), file_name.to_owned())];
-    let extra_entries = parser.get_entry_points(&document, language);
+    let mut entries = parser.get_entry_points(&document, language);
 
-    entries.extend(
-        extra_entries
-            .iter()
-            .map(|(e, p)| (Some(&e[..]), PathBuf::from((*p).to_owned() + ".temp"))),
-    );
+    let file_name_without_ext = file_name.with_extension("");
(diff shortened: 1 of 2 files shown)


mlange-42:
I will have to check this more in-depth after dinner. Also, not really sure what the parent thing was for and what might change by removing it.

adamreichold:
> I will have to check this more in-depth after dinner.

I think I understand the old code better now. The original file name like `README.md.md` got the last `.md` removed by calling `.file_stem()`. So to make things work out when using `//- file:main.rs`, those paths were extend to `main.rs.temp` to have second extension to cut off using `.file_stem()`.

In this version, I just turned things around: I call `.with_extension("")` on the original file name to go from `.ext.md` to `.ext` but then leave all other sub file names alone avoid the extra work.

> Also, not really sure what the parent thing was for and what might change by removing it.

The incantation
```
if let Some(parent) = file_name.parent() {
  file_path.push(parent);
}
file_path.push(file_name.file_stem().unwrap());
```
is basically equivalent to
```rust
let file_name = file_name.with_extension("");
file_path.push(file_name);
```
but avoids the additional allocation implied by `Path::with_extension`, i.e. it adds all of `file_name` to `file_path`, but without the last extension of `file_name`. (But then in this case, we did two extra allocations to be able to call `Path::file_stem` in the first place...)

mlange-42:
Ah... now I get it, too. So I wrote the convoluted parent thing simply because I was not aware of `with_extension` to get the file stem incl. its path...

mlange-42 (approved):

PR #64, opened 2021-01-30
Move path mangling to util module and test it.

adamreichold:
This might be leaning a bit into gamification, but it does drop one more temporary allocation and sets a bit of precedent towards better testing...


mlange-42 (approved):

mlange-42:
I think more tests in general would be healthy. Therefore 👍

PR #65, opened 2021-01-30
Reworking project initialization

mlange-42:
* [x] Rename `create` to `init`
* [x] Remove argument `main_file`
* [x] Rework template to produce a minimal Rust project
* [x] Adapt README/book
* [x] Adapt examples


mlange-42:
Hm... not so easy to adapt the `Blocks and macros` chapter for the new `init` result...

examples/features_demo/README.md
```diff
@@ -2,12 +2,66 @@
 
 This examples demonstrates Yarner's most important features by creating a simple but complete [Rust](https://rust-lang.org) project.
```
adamreichold, line 3: I guess the overlap between the default setup and this one is by now quite high? What do you think about having `examples/rust` and `examples/java` showing how to do a small Gradle project instead to highlight that Yarner is not married to Rust. (And of course, `examples/netlogo` if you feel adventurous...) Personally, I think the book is a better place to ensure feature coverage as I prefer examples to are easy to digest.

src/create/Yarner.toml
```diff
@@ -55,10 +54,10 @@ code = "code/"
 docs = "docs/"
 
 # The input source file(s) as glob pattern(s).
-files = ["%%MAIN_FILE%%"]
+files = ["README.md"]
 
 ## Entrypoint block name. Optional. If not supplied, unnamed code blocks are used.
-# entrypoint = "Main"
+#entrypoint = "Main"
```
adamreichold, line 60: I think the other commented-out defaults have a space after the hash?

src/create/mod.rs
```diff
@@ -40,4 +38,4 @@ impl Drop for RemoveOnDrop<'_> {
 
 const CONFIG: &str = include_str!("Yarner.toml");
 
-const DOCUMENT: &str = include_str!("document.md");
+const DOCUMENT: &str = include_str!("README.md");
```
adamreichold, line 41: Since we are not replacing anything anymore, we can use `include_bytes!` here and drop the `.as_bytes()` call above.

examples/features_demo/code/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "hello-world"
+name = "features-demo"
```
adamreichold, line 2: Not sure if we should really commit the lock files?

mlange-42: Already thinking about ignoring outputs completely. Although I would like to have the examples complete without requiring installation of yarner, but it bothers me that rust binary examples complicate installation.

examples/features_demo/README.md
mlange-42: Too bad that reverse mode is not possible with NetLogo, as it allows for comments only in the code part, but not in the rest. Using `.nls` files could be a solution, however.

mlange-42: Also, I realize that there should be a way to define language settings while still having no block labels in the output, which is cuttently not possible.

mlange-42: See #68 and #67

examples/features_demo/code/Cargo.toml
mlange-42: See #69

examples/features_demo/README.md
mlange-42: Regarding the examples, do you mean to have a single example for each language, or more like having e.g. the two rust examples in a sub-folder `rust`?

examples/features_demo/code/Cargo.toml
adamreichold: > it bothers me that rust binary examples complicate installation.

This is a non-issue once we publish to crates.io...

examples/features_demo/README.md
adamreichold: No, I meant a single example for each language which keeps it simple so it is easy to understand.

PR #71, opened 2021-01-31
Rework of examples

### LINKED ISSUE #68: [Examples] Add examples for other languages than Rust (Java, NetLogo)

### LINKED ISSUE #69: [Examples] Ignore output folders of examples?

mlange-42:
* [x] Ignore output of example
* [x] Simplify installation instructions
* [x] Add an example for Java and Gradle
* [x] Add a NetLogo example
* [x] Add a Python example

Fixes #68
Fixes #69


adamreichold:
I would be glad if we could merge the default_setup and features_demo examples into a single rust example.

mlange-42:
As `default_setup` is just what is created with `init`, I would simply delete it. What do you thinkabout activating the reverse mode in the then `rust` example, and describe it like in the netlogo example (except for the netlogo complications)?

mlange-42:
Done. Moved remarks on dreverse mode in the netlogo example to the borrom

adamreichold:
> What do you thinkabout activating the reverse mode in the then rust example

Yes, I think reverse mode makes most sense for languages like Rust and Java which provide good native tooling like formatters and IDE which one would want to use, but which cannot handle the literate program.

adamreichold:
(A proper review on my part will not happen before tomorrow, but don't let yourself be held up if you are in a hurry.)

mlange-42:
Ok, thanks for letting me know. I will see. I case I venture to give the restructuring that you suggested a try, I would prefer to utilize your capacity for that.

PR #73, opened 2021-01-31
Reverse mode fixes for multiple occurrences of a code block

### LINKED ISSUE #72: [Bug] Handle multiple occurences of the same code block in reverse mode
It is perfectly legal to use the same code block through macros in multiple places. In reverse mode, one of the blocks may have changed while another may not. It is ambiguous which block tu use in this case.

Also, it would currently be impossible to determine whether multiple blocks of the same name were concatenated, or whether it is a repetition of the same block.

Possible solutions:

* Abort with error
* Switch to interactive mode and let the user select

mlange-42:
Write code block index (for identifying concatenated blocks) to block labels, and also derive from them. This allows to detect "real" multiple occurrences of a block.

Currently, we warn on multiple equal occurrences, and abort with an error on differing occurences.
Should the user be allowed to decide which block to use?

Fixes #72

### DIFF
--- a/guide/src/config-languages.md
+++ b/guide/src/config-languages.md
@@ -38,3 +38,4 @@ but block labels in the target language are not wanted or not supported, leave o
 | `block_start`       | Start sequence of block labels                                                                                                      |
 | `block_next`        | Start of next block with the same name                                                                                              |
 | `block_end`         | End of block labels                                                                                                                 |
+| `name_separator`    | Separator between block source file, name and index (e.g. `// <@main.rs#A clode block#0`)                                                  |
--- a/guide/src/config-languages.md
+++ b/guide/src/config-languages.md
@@ -21,6 +21,7 @@ eof_newline = true
     block_start = "<@"
     block_next = "<@>"
     block_end = "@>"
+    name_separator = "#"
 ```
 
 When language settings are requires to set `clear_blank_lines` or `eof_newline`,
--- a/guide/src/config-languages.md
+++ b/guide/src/config-languages.md
@@ -21,7 +21,6 @@ eof_newline = true
     block_start = "<@"
     block_next = "<@>"
     block_end = "@>"
-    name_separator = "#"
 ```
 
 When language settings are requires to set `clear_blank_lines` or `eof_newline`,
@@ -38,4 +37,3 @@ but block labels in the target language are not wanted or not supported, leave o
 | `block_start`       | Start sequence of block labels                                                                                                      |
 | `block_next`        | Start of next block with the same name                                                                                              |
 | `block_end`         | End of block labels                                                                                                                 |
-| `name_separator`    | Separator between block source file, name and index (e.g. `// <@main.rs#A clode block#0`)                                                  |
--- a/src/document/code.rs
+++ b/src/document/code.rs
@@ -196,6 +196,10 @@ impl Line {
             .and_then(|s| s.block_labels.as_ref())
             .map(|l| &l.block_next[..])
             .unwrap_or("");
+        let block_name_sep = settings
+            .and_then(|s| s.block_labels.as_ref())
+            .map(|l| &l.name_separator[..])
+            .unwrap_or("");
 
         let clean = if let Some(s) = settings {
             s.clean_code || s.block_labels.is_none()
@@ -247,11 +251,12 @@ impl Line {
 
                     if !clean {
                         result.push(format!(
-                            "{}{} {}{}#{}{}",
+                            "{}{} {}{}{}{}{}",
                             &self.indent,
                             comment_start,
                             if idx == 0 { &block_start } else { &block_next },
                             path,
+                            block_name_sep,
                             name,
                             comment_end,
                         ));
@@ -275,8 +280,14 @@ impl Line {
 
                     if !clean && idx == blocks.len() - 1 {
                         result.push(format!(
-                            "{}{} {}{}#{}{}",
-                            &self.indent, comment_start, &block_end, path, name, comment_end,
+                            "{}{} {}{}{}{}{}",
+                            &self.indent,
+                            comment_start,
+                            &block_end,
+                            path,
+                            block_name_sep,
+                            name,
+                            comment_end,
                         ));
                     }
                 }
--- a/src/document/code.rs
+++ b/src/document/code.rs
@@ -300,6 +300,7 @@ impl Line {
                         .unwrap();
                     }
                 }
+                result.pop();
                 Ok(result)
             }
         }
--- a/src/document/code.rs
+++ b/src/document/code.rs
@@ -251,13 +251,15 @@ impl Line {
 
                     if !clean {
                         result.push(format!(
-                            "{}{} {}{}{}{}{}",
+                            "{}{} {}{}{}{}{}{}{}",
                             &self.indent,
                             comment_start,
                             if idx == 0 { &block_start } else { &block_next },
                             path,
                             block_name_sep,
                             name,
+                            block_name_sep,
+                            idx,
                             comment_end,
                         ));
                     }
@@ -280,13 +282,15 @@ impl Line {
 
                     if !clean && idx == blocks.len() - 1 {
                         result.push(format!(
-                            "{}{} {}{}{}{}{}",
+                            "{}{} {}{}{}{}{}{}{}",
                             &self.indent,
                             comment_start,
                             &block_end,
                             path,
                             block_name_sep,
                             name,
+                            block_name_sep,
+                            idx,
                             comment_end,
                         ));
                     }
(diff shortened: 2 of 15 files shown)


src/document/mod.rs
```diff
@@ -103,14 +104,11 @@ impl Document {
...
-                        blocks.get(index)
-                    } else {
-                        None
-                    };
+                    let alt_block = code_blocks.get(&(&code_block.name, &index));
+
                     output.push_str(
                         &printer
-                            .print_code_block_reverse(code_block, alt_block)
+                            .print_code_block_reverse(code_block, alt_block.copied())
```
mlange-42, line 111: Not really sure how to best handle (or avoid) the `Option<&&RevCodeBlock>` here.

adamreichold:
> Should the user be allowed to decide which block to use?

I think an option to force using the first one or maybe the one that differs from the source if it is only one might be practical, but interactive selection sounds a bit too much.

adamreichold:
This is me being a bit late to the party: But why is reverse mode a flag instead of a subcommand?

guide/src/config-languages.md
```diff
@@ -37,3 +38,4 @@ but block labels in the target language are not wanted or not supported, leave o
 | `block_start`       | Start sequence of block labels                                                                                                      |
 | `block_next`        | Start of next block with the same name                                                                                              |
 | `block_end`         | End of block labels                                                                                                                 |
+| `name_separator`    | Separator between block source file, name and index (e.g. `// <@main.rs#A clode block#0`)                                                  |
```
adamreichold, line 41: If this necessarily part of a comment in the target language, couldn't Yarner select this automatically? The only requirement seems to be that it is not part of a code block name and even this could be worked around using escaping.

src/document/code.rs
```diff
@@ -196,6 +196,10 @@ impl Line {
             .and_then(|s| s.block_labels.as_ref())
             .map(|l| &l.block_next[..])
             .unwrap_or("");
+        let block_name_sep = settings
+            .and_then(|s| s.block_labels.as_ref())
```
adamreichold, line 200: I think
```rust
.and_then(|s| s.block_labels.as_ref())
.map(|l| &l.name_separator[..])
.unwrap_or("");
```
could be replaced by
```
.and_then(|s| s.block_labels.as_deref())
.unwrap_or_default();
```

(Generally speaking, I would prefer `foo.as_str()` over `&foo[..]` as the intent is easier to read.)

guide/src/config-languages.md
mlange-42: That is how it was before. I thought the added flexibility would be better as when hardcoded it may collide with characters in names, or with block comment end sequences (for the rare case of languages that support no line comments).

src/document/code.rs
```diff
@@ -247,12 +251,15 @@ impl Line {
 
                     if !clean {
                         result.push(format!(
```
adamreichold, line 253: I understand this was not introduced in this MR, but this function basically does
```rust
let mut result = Vec::new();
...
result.push(format!("...", ...));
...
Ok(result.join("\n"))
```
among other things. Using `write!` like
```
let mut result = String::new();
...
writeln!(&mut result, "...", ...).unwrap();
...
Ok(result)
```
should significantly more efficient.

src/document/mod.rs
```diff
@@ -9,6 +9,7 @@ use self::transclusion::Transclusion;
 
 use crate::config::LanguageSettings;
 use crate::parser::{code::RevCodeBlock, md::MdParser};
+use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::hash_map::{Entry, HashMap};
```
adamreichold, line 13: You can collapse these to
```rust
use std::collections::hash_map::{HashMap, Entry::{self, Occupied, Vacant}};
```

src/document/mod.rs
```diff
@@ -166,7 +168,16 @@ impl Document {
         let mut result = String::new();
         match code_blocks.get(&entrypoint) {
             Some(blocks) => {
+                let mut block_count: HashMap<&Option<String>, usize> = HashMap::new();
                 for (idx, block) in blocks.iter().enumerate() {
+                    let index = match block_count.entry(&block.name) {
```
adamreichold, line 173: I think something like
```
let count = block_count.entry(&block.name).or_default();
*count += 1;
count - 1
```
might be a bit easier to read.

adamreichold: (In any case, the `Occupied` branch should read `*entry.get_mut() += 1;`)

src/parser/code.rs
```diff
@@ -45,15 +43,14 @@ impl CodeParser {
         source: &str,
         parser: &MdParser,
         block_labels: &BlockLabels,
-    ) -> Vec<RevCodeBlock> {
+    ) -> Result<Vec<RevCodeBlock>, String> {
```
adamreichold, line 46: I think going for `Fallible` here would be the future proof choice.

src/parser/code.rs
```diff
@@ -88,6 +85,10 @@ impl CodeParser {
                         Some(s.to_string())
                     }
                 });
+                let index_str = parts.next().unwrap_or("0");
```
adamreichold, line 88: Why the `unwrap_or` call instead of failing?

guide/src/config-languages.md
adamreichold: Hhhmmm, I think flexibility has a price mostly w.r.t. the mental energy required to make the abstract multitude of options concrete in ones mind, so I would also suggest weighing the cost against the benefits of added flexibility/configurability.

In this particular case, I think I would have stuck to '#' until someone actually has a problem with this choice.

Actually checking that block labels do not contain this character (where fixed or configurable) seems worthwhile though...

mlange-42:
> This is me being a bit late to the party: But why is reverse mode a flag instead of a subcommand?

Because all the arguments for the "forward" mode can also be used.

guide/src/config-languages.md
mlange-42: > In this particular case, I think I would have stuck to '#' until someone actually has a problem with this choice.

Maybe this is the best way do go for now.

mlange-42:
Meanwhole, I am a bit afraid of conflicts with the other MR on top of this one. Although it should probably be rebased after merging this one, maybe I should merge the other one first?

adamreichold:
> Because all the arguments for the "forward" mode can also be used.

I think since the forward mode is not a subcommand, its arguments are inherited by all subcommands anyway. (E.g. `init` incorrectly accepts all forward mode flags via
```
> yarner --some-flag init
```
as well.)

> maybe I should merge the other one first?

As these MR are meant to reduce the context to enable a sufficiently thorough review, I would definitely prefer a rebase to merging just to avoid conflicts.

mlange-42:
I think I am convinced now to remove meta variables. Thus, the other MR is mostly obsolete. I will fix this one now, and then go for removing meta vars.

src/document/code.rs
mlange-42: So you mean constructing an entire default `BlockLabels` struct?

adamreichold: > So you mean constructing an entire default BlockLabels struct?


No, I just meant replacing `.map(|l| &l.name_separator[..])` by `.as_deref()` to get from `&Option<String>` to `Option<&str>` and replacing `.unwrap_or("")` by the more speaking `unwrap_or_default()`.

mlange-42: `name_sep` is gone now, but it still applies to `comment_end`

src/document/mod.rs
mlange-42: With the last version, I get no return value (which is the old value, before `insert`). But I will try to understand the first verion, first.

adamreichold: Both `or_default()` and `get_mut()` yield a `&mut usize` which can be used to both modify the value in-place and query its current value by just dereferencing. The variant the "returns" `*count - 1` is just avoid doing something like
```
let old_count = *count;
*count += 1;
old_count
```
because the compiler will transform `count - 1` into something like this anyway.

mlange-42: Ok, cool! also changed in another place where it was expressed even more complicated.

src/parser/code.rs
mlange-42: At this point, I was not really sure how it would look like in the end, and didn't revisit it. Thanks!

mlange-42:
Made reverse mode a subcommand

src/document/code.rs
mlange-42: Hm, It is not completely equivalent, as now there is a newline at the end of the line...

adamreichold: You can do `result.pop()` before returning if that is an issue.

mlange-42: Currently trying exactly that, but with no effect on the result. Maybe the problem is elsewhere, introduced in the same commit.

mlange-42: Ah, no, my fault... I was playing with the sources already "broken" by a reverse run. Fixed...

PR #75, opened 2021-02-01
Meta variables as optional feature

### LINKED ISSUE #74: [Feature] Make meta variables a feature to be activated explicitly, as it collides with reverse mode.
Meta variables should not be allowed by default.

Make it a feature to be explicitly activated, e.g. by wrapping the three related `toml` options into an optional struct.
Should we then detect if it is enabled and any language is set up for reverse mode, and emit a warning?

Also mention the limitation is the guide.

mlange-42:
* [x] Make meta variables an optional feature
* [x] Adapt examples
* [x] Adapt user guide

Fixes #74

### DIFF
--- a/guide/src/meta-variables.md
+++ b/guide/src/meta-variables.md
@@ -1,8 +1,10 @@
 # Meta variables
 
+If you consider a macro invocation like a function call, then meta variables are like parameters.
+
 [[_TOC_]]
 
-If you consider a macro invocation like a function call, then meta variables are like parameters.
+## Meta variables
 
 By default, to indicate that a macro includes a meta variable, the name of the variable must be part of the name of the macro, delimited by `@{` and `}`.
 Then that meta variable may be used within the macro by again using its name within the `@{` and `}` in the code.
@@ -28,6 +30,8 @@ Now, to say things to many people:
 ```
 ````
 
+## Default values
+
 Meta variables can have default values:
 
 ````markdown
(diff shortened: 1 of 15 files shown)


guide/src/meta-variables.md
```diff
@@ -1,8 +1,10 @@
 # Meta variables
 
+If you consider a macro invocation like a function call, then meta variables are like parameters.
+
 [[_TOC_]]
 
-If you consider a macro invocation like a function call, then meta variables are like parameters.
+## Meta variables
```
adamreichold, line 7: I guess I am already on the record that I think meta variables definitely overdo it and should be removed entirely as this kind of abstraction is what the programming languages themselves should be used for?

src/config.rs
```diff
@@ -26,15 +26,26 @@ impl Config {
...
         for language in self.language.values() {
             language.check()?;
+            if language.block_labels.is_some() {
+                any_reverse = true;
+            }
+        }
+
+        if any_reverse && self.parser.meta_variables.is_some() {
+            return Err(
+                "Reverse mode can't be combined with the meta variables feature.\n\
```
adamreichold, line 44: Personally, I think text spanning multiple lines in the source is more readable when embedded as raw string literals.

guide/src/meta-variables.md
mlange-42: Yes, you are 😉 

Maybe the fact that I never used them so far should convince me to really remove them now... (although not used Yarner in a ton of projects...)

mlange-42:
Decided to remove the mate variables feature completely. See #77

guide/src/meta-variables.md
DPDmancul: Metavariables are very useful to literate configurations: usually configuration files don't allow loops and having the metavariables feature could be very helpful to avoid rewriting the same thing many times changing only one piece.

PR #77, opened 2021-02-01
Remove the meta variables feature

mlange-42:


mlange-42:
@adamreichold Ok, I was wrong whan I said I never used meta variables. They were essential to build NetLogo UI. The UI syntax in NetLogo files is almost impossible to handle without them. Hm, that really makes me doubt...

adamreichold:
> They were essential to build NetLogo UI. The UI syntax in NetLogo files is almost impossible to handle without them. Hm, that really makes me doubt...

I remember that discussion. I would not subscribe to "essential" though. I would rather use "convenient" as they would avoid a lot of repetitive code. I think my main issue at the time was that a literate programming tool should not be used to work around the deficiencies of the target language.

What do people building plain NetLogo models use to handle this? Does NetLogo provide any RAD-like editors for the UI parts? Is that work flow compatible with Yarner's reverse mode? If so, I would argue that using NetLogo's native design tools to basically ingest the UI code as a opaque blob, probably even hidden in the documentation output would actually be preferable than trying to make writing this stuff by hand more manageable using macros and meta variables.

mlange-42:
> I would rather use "convenient" as they would avoid a lot of repetitive code.

The problem is not the repetitive code, but the format / "syntax". Take a look at the example and you will see...

> What do people building plain NetLogo models use to handle this?

NetLogo has a dedicated "Tab" for UI etc, and normally everything is created manually, probably almost never from code. Have you never seen NetLogo?

Reverse mode is compatibel with NetLogo when the logic is moved into a `.nls` file, as in the example. That is because NetLogo does not support comments, exept in the code part.

mlange-42:
I also re-opened the other MR regarding this topic, with a rebased branch. #78

mlange-42:
> If so, I would argue that using NetLogo's native design tools to basically ingest the UI code as a opaque blob, probably even hidden in the documentation output would actually be preferable than trying to make writing this stuff by hand more manageable using macros and meta variables.

Maybe this is indeed an option: Mave the NetLogo file with only UI and the other ugly stuff, as well as `__includes["Codenls"]` in the code tab, as a copy-code file. One would then only LP the `.nls` file, and code-copy the `.nlogo`.

With reverse mode, it should then even be possible to edit the UI in the code output, as the reverse mode also works for the copy-code files (it copies them back).

🤔

mlange-42:
I reworked the NetLogo model to give it a try.
With reverse mode, it indeed works very well:

1) Start with a minimal `.nlogo` file in a sub-directory to be copied. Put `__includes["Code.nls"]` as only code in the code tab
2) Start LP the actual code
3) Run `yarner`
4) Open the `.nlogo` **in the generated code output**, edit UI
5) Run `yarner reverse`
6) Continue to LP the actual code

examples/netlogo/README.md
```diff
@@ -135,3 +135,8 @@ The content of the code tab is created by simply drawing together all blocks sho
 ; ==> Go.
 ; ==> Submodels.
 ```
+
+### NetLogo file
+
+In the main `nlogo` file, we only "include" an `nls` file to allow for the reverse mode.
```
adamreichold, line 141: Maybe explicitly mention that this also enables using NetLogo's RAD tools to modify the project and its UI?

src/document/code.rs
```diff
@@ -266,20 +208,18 @@ impl Line {
...
-                                            "".to_string()
-                                        } else {
-                                            format!("{}{}", self.indent, line)
-                                        }
-                                    })
-                                    .collect::<Vec<_>>()
-                                    .join("\n")
-                            })?,
+                        block.compile_with(code_blocks, settings).map(|code| {
+                            code.split('\n')
```
adamreichold, line 212: Not sure if you want to touch this now or just earmark it for later, but
* `some_str.split('\n')` is probably better written as `some_str.lines()` using [`str::lines`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.lines)
* this might benefit from using the non-collecting `JoinExt::join` function or even manually writing everything into a single `String` to avoid the intermediate `format!` calls for the indentation

src/parser/md.rs
```diff
@@ -290,25 +256,16 @@ impl MdParser {
...
-                                            block.name = Some(name.to_string());
-                                            block.hidden = hidden;
-                                            block.vars = vars;
-                                            block.defaults = defaults;
-                                        }
-                                        Err(error) => {
-                                            return Some(Parse::error(error, line_number));
-                                        }
+                                    let name = line.trim()[self.comment_start.len()..].trim();
+                                    let hidden = name.starts_with(&self.hidden_prefix);
```
adamreichold, line 260: This would probably be a good place to use [`str::strip_prefix`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.strip_prefix)

src/parser/md.rs
```diff
@@ -484,44 +395,20 @@ impl MdParser {
 
         if rest.starts_with(&self.macro_start) {
```
adamreichold, line 396: Another good candidate for `strip_prefix` I guess.

adamreichold:
LGTM after the NetLogo clarification. The other stuff is clean-up and not really related to the issue at hand.

src/document/code.rs
mlange-42: Done. Besides avoiding collect and format, it also makes the code much more concise:

```rust
let code = block.compile_with(code_blocks, settings)?;
for line in code.lines() {
    if blank_lines && line.trim().is_empty() {
        writeln!(result).unwrap();
    } else {
        write!(result, "{}", self.indent).unwrap();
        writeln!(result, "{}", line).unwrap();
    }
}
```

Maybe an alternative to `line.trim().is_empty()` would be helpful, but is does not allocate, so it is probably ok.

src/parser/md.rs
mlange-42: Did it like this, but not sure if that's the solution you had in mind:

```rust
let mut name = line.trim()[self.comment_start.len()..].trim();
let hidden = if let Some(stripped) = name.strip_prefix(&self.hidden_prefix) {
    name = stripped;
    true
} else {
    false
};
```

examples/netlogo/README.md
mlange-42: Added this at the very end:

This separation of model code and user interface also allows to edit the model's UI elements in NetLogo's GUI builder tool, while using Literate Programming for the code.

src/parser/md.rs
mlange-42: And even to combine it with `strip_suffix` 😉

src/parser/md.rs
adamreichold: I think we can avoid the temporaries, e.g.
```rust
block.name = line.trim()[self.comment_start.len()..].trim();
if let Some(stripped) = name.strip_prefix(&self.hidden_prefix)
{
    block.name = stripped;
    block.hidden = true;
}
```
and there is another instance of the strip-prefix pattern above concerning `comment_start`.

mlange-42: Ok, thanks. Will do that before the next merge.

mlange-42: > and there is another instance of the strip-prefix pattern above concerning comment_start.

Too bad that `if` and `if let` can't be combined...

adamreichold: This is under consideration but it looks a funky admittedly. You can achieve a similar effect using guard pattern even now though:
```rust
match foo {
  Some(foo) if foo.baz && qux => { ... }
  _ => (),
}
```

mlange-42: Ok, but that makes it complicated again due to necessity for trimming. I'll leave the outer `if` as is, but changed to avoid the temporaries (although a bit more compicated, as block.name is an Option):

```rust
if let Some(stripped) = name.strip_prefix(&self.hidden_prefix) {
    block.name = Some(stripped.to_string());
    block.hidden = true;
} else {
    block.name = Some(name.to_string());
};
```

PR #78, opened 2021-02-01
[Withdrawn] Meta variables as optional feature

### LINKED ISSUE #74: [Feature] Make meta variables a feature to be activated explicitly, as it collides with reverse mode.

mlange-42:
**This will probably not be merged, in favor of #77**

*Just letting it here for a bit for comparison and to let the decision settle.*

* [x] Make meta variables an optional feature
* [x] Adapt examples
* [x] Adapt user guide

Fixes #74

### DIFF
--- a/examples/java_gradle/Yarner.toml
+++ b/examples/java_gradle/Yarner.toml
@@ -16,9 +16,10 @@ link_prefix = "@"
 file_prefix = "file:"
 hidden_prefix = "hidden:"
 
-interpolation_start = "@{"
-interpolation_end = "}"
-variable_sep = ":"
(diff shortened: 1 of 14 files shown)


mlange-42:
I realized that they are indeed useful. Particularly to create NetLogo UI elements, they are almost indispensable (see the NetLogo example's appendix).

PR #79, opened 2021-02-02
Temporarily disabled comment extraction, adapted guide and examples.

mlange-42:


src/parser/md.rs
```diff
@@ -47,14 +47,18 @@ pub struct MdParser {
...
     pub fence_sequence_alt: String,
+    /// The language to set if there was no automatically detected language. Optional
+    pub default_language: Option<String>,
+    /// Temporary switch to disable comment extraction
+    #[serde(skip)]
+    pub allow_comment_extraction: bool,
     /// Parsed comments are stripped from the code and written to an `<aside></aside>` block after
     /// the code when printing. If false, the comments are just written back into the code.
+    #[serde(skip)]
     pub comments_as_aside: bool,
```
adamreichold, line 58: If I understand this correctly, this means that a `Yarner.toml` containing this cannot be parsed anymore? If this is intended I would suggest just removing the feature completely (and resurrecting the code from Git when reimplementing it).

Another viable approach might be to continue parsing the field but display a warning or fail with an error if it is enabled?

mlange-42: Yes, I added a switch to disable/enable comment extraction (to be more exact, to identify them during parsing), and used the serde attribute to force the two related options to `false`.

There are several places where comments are handled. I think they don't need a lot of modification when re-enabled, although with different syntax. So I thought it is the best option to simply "switch if off" to have the code there for experimentation.

adamreichold: > and used the serde attribute to force the two related options to false.

But doesn't this also mean that they cannot be parsed at all? I think avoiding `#[serde(skip)]` and raising an explicit error during initialisation would be preferable from a usability point of view.

mlange-42: Yes, they are completely disabled currently. I could add `#[serde(default)]` instead of `#[serde(skip)]` and throw an error if one of them is set to `true`.

adamreichold: Yes, I think that would be preferable as the error message could say something like "currently disabled but will be available soon, just comment out the field in your toml until the next version".

PR #86, opened 2021-02-03
Track code files over all input patterns and prevent multiple writes

mlange-42:
With this in the `toml`, it was possible to write multiple times to the same code file (i.e. once per entry):

```
files = ["file-1.md", "file-2.md"]
```

Further modified to only prevent multiple writes if they really come from different source files (respecting transclusion sources)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -362,6 +362,7 @@ fn process_inputs_forward(
     language: Option<&str>,
 ) -> Result<(), String> {
     let mut any_input = false;
+    let mut track_code_files = HashSet::new();
     for pattern in input_patterns {
         let paths = match glob::glob(&pattern) {
             Ok(p) => p,
@@ -407,7 +408,7 @@ fn process_inputs_forward(
                     language,
                     &config.language,
                     &mut HashSet::new(),
-                    &mut HashSet::new(),
+                    &mut track_code_files,
                 ) {
                     return Err(format!(
                         "Failed to compile source file \"{}\": {}",
--- a/src/main.rs
+++ b/src/main.rs
@@ -362,7 +362,7 @@ fn process_inputs_forward(
     language: Option<&str>,
 ) -> Result<(), String> {
     let mut any_input = false;
-    let mut track_code_files = HashSet::new();
+    let mut track_code_files = HashMap::new();
     for pattern in input_patterns {
         let paths = match glob::glob(&pattern) {
             Ok(p) => p,
@@ -594,7 +594,7 @@ fn compile_all(
     language: Option<&str>,
     settings: &HashMap<String, LanguageSettings>,
     track_input_files: &mut HashSet<PathBuf>,
-    track_code_files: &mut HashSet<PathBuf>,
+    track_code_files: &mut HashMap<PathBuf, Option<PathBuf>>,
 ) -> Fallible {
     if !track_input_files.contains(file_name) {
         let mut document = transclude(parser, file_name)?;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -708,14 +709,17 @@ fn compile(
...
 ) -> Fallible {
     eprintln!("Compiling file {}", file_name.display());
 
     let mut entries = parser.get_entry_points(&document, language);
 
     let file_name_without_ext = file_name.with_extension("");
-    entries.insert(entrypoint, &file_name_without_ext);
+    entries.insert(
+        entrypoint,
+        (&file_name_without_ext, Some(PathBuf::from(file_name))),
```
adamreichold, line 721: Wouldn't `file_name.to_owned()` do the same yet better capture the intent?

src/parser/md.rs
```diff
@@ -428,13 +428,23 @@ Please comment out option `comments_as_aside` until the next version, and rename
...
         let pref = &self.file_prefix;
         for block in doc.code_blocks(language) {
             if let Some(name) = block.name.as_deref() {
                 if let Some(rest) = name.strip_prefix(pref) {
-                    entries.insert(Some(name), Path::new(rest));
+                    entries.insert(
+                        Some(name),
+                        (
+                            Path::new(rest),
+                            if let Some(f) = &block.source_file {
```
adamreichold, line 441: This looks like it wants to be
```rust
block.source_file.as_ref().map(|file| file.into())
```
or something similar?

src/main.rs
```diff
@@ -745,14 +749,22 @@ fn compile(
...
-                        "Multiple locations point to code file {}",
-                        file_path.display()
-                    )
-                    .into());
-                } else {
-                    track_code_files.insert(file_path.clone());
+                match track_code_files.entry(file_path.clone()) {
+                    Occupied(entry) => {
+                        if sub_source_file == *entry.get() {
+                            eprintln!("  Skipping file {} (already written)", file_path.display());
```
adamreichold, line 755: Shouldn't informational output go to stdout?

src/main.rs
```diff
@@ -745,14 +749,22 @@ fn compile(
...
+                            continue;
+                        } else {
+                            return Err(format!(
+                                "Multiple locations point to code file {}",
```
adamreichold, line 759: Maybe extend this to "multiple distinct locations"?

src/main.rs
```diff
@@ -807,13 +822,7 @@ fn compile_reverse(
...
                 }
 
-                if track_code_files.contains(&file_path) {
-                    return Err(format!(
-                        "Multiple locations point to code file {}",
-                        file_path.display()
-                    )
-                    .into());
-                } else {
+                if !track_code_files.contains(&file_path) {
```
adamreichold, line 825: The `contains` call seems redundant as `insert` returns if the value was already present.

src/main.rs
mlange-42: Thanks! Was not aware of `Path::to_owned() -> PathBuf`

src/main.rs
mlange-42: Currently, all informational output is done via `eprintln!`, as I thought stdout shoulde be reserved for machine-readable output. However, there is no such output in the current version anymore (in an older version treating only a single file. there was).

src/main.rs
mlange-42: Indeed... forgot that when removing the check/error that is not necessary in reverse mode

src/main.rs
adamreichold: > as I thought stdout shoulde be reserved for machine-readable output.

Yes, if you application is able to produce "results" via stdout, then all human-targeted information should go to stderr. But since Yarner does not do that, the stdout/stderr split can be used for informational/error signalling instead. Both are just conventions though, so pushing everything via stderr does work as well. It just leaves some signalling potential "on the table."

mlange-42: If we really add the possibility for pre/postprocessors, we wil definitely have machine-readable output.

adamreichold: But that output will probably not be transferred via our stdout. If the processors are separate binaries, we will most likely spawn child processes and exchange data via _their_ stdin/stdout, but ours would stay firmly targeted at human consumption.

mlange-42: So you suggest to print everything except error messages to stdout?

adamreichold: Yes, I would find that reasonable.

Issue #57, opened 2021-01-30
[Design] Task-oriented module structure


adamreichold:
I have tried to reduce clutter and simplify technicalities so far, but I think with the current status (no more library-binary split, support only for Markdown), we could reach a more code organisation by splitting our modules along the tasks, e.g.

* `parse`: Parse Markdown in documents a.k.a. list of blocks. Contains both the data structure as well as the parser producing it similar to `parser::code` now.
* `print`: Prints documents based on the documents created above.
* `compile`: Compiles source files based on the documents created above.
* `reverse_compile`: Merge source files into code blocks.

I am mainly hoping to be able to move a lot of code from the main module into these task-specific sub modules. But I also hope to reduce the "code scatter", e.g. unrelated functionality being in a single module like `document` as well as related functionality being scattered over multiple modules like compiling which mostly happens in the main module but also within `document::code`.

@mlange-42 What do you think?

mlange-42:
This sounds like a very useful measure! I also think the structure you propose sounds reasonable.

In line with the restructuring, we may already prepare for what I proposed yesterday in #53, in case you would support that idea. However, when returning to the binary/library split for that sake, we probably don't need to expose much more than `Document` and its content structs.

adamreichold:
I think #53 does not imply a library/binary split as those external processors would not need to be Rust programs at all. The stable interface for the external processors would be a [JSON schema](https://crates.io/crates/schemars), not a Rust library.

We could factor out a Rust library that contains really only the serializable structures so that processor writers do not need to do the parsing themselves. But if our data model is so complicated as to need this, we should probably provide the same courtesy to all implementors and provide a proper schema, e.g. a Protobuf specification, that can be used from any language and provides more facilities for schema evolution.

mlange-42:
Ok, that sounds reasonable. No real experience in this field so far on my side...

mlange-42:
Done in #89, #98, #99, #102, #103, #104.

PR #87, opened 2021-02-04
[Withdrawn] Task-oriented module structure

mlange-42:
*[Withdrawn] Still here for comparison with step-wise re-structuring.*

A trial to restructure according #57.

So far:

* Moved parser and document functions to modules `parse` and `print`
* Moved compile-related functions from `main` to modules `compile` and `compile_reverse`
* Removed struct `CodeParser`, moved functionality to module `code`

@adamreichold Any advice, or an independent trial, is highly appreciated.


adamreichold:
> Any advice, or an independent trial, is highly appreciated.

Instead of a duplicate approach, I would prefer making this easier to review by breaking it up. For example, create on new module per pull request. Or even smaller steps. (Of course, I can look at individual commits here, but the discussion should be scoped as well. And since you already have individual commits, it should also be straight forward to break this into independent pull requests.)

mlange-42:
I tried to break it into commits, but it was not possible to do it module by module. I can, however, start with a MR for the first commit.

mlange-42:
My impression was that this is basically only moving around code, so the overall structure changes but content not. So I thought you would primarily focus on the overall structure rather than diffs, anyway. But maybe more places for improvement become obvious by doing it step by step.

adamreichold:
> But maybe more places for improvement become obvious by doing it step by step.

Often, doing larger reshuffling of a code base is also used to tease out other improvements since you are reading through all of it anyway.

mlange-42:
Done in #89, #98, #99, #102, #103, #104.

PR #89, opened 2021-02-04
Move parser and document functions to modules print and parse

mlange-42:


src/document/mod.rs
```diff
@@ -327,6 +144,30 @@ impl Document {
             index += 1;
         }
     }
+
+    /// Finds all file-specific entry points
+    pub fn get_entry_points<'a>(
```
adamreichold, line 149: `get_` is a bit uncustomary for Rust. I think just `entry_points` would be fine.

src/parse.rs
```diff
@@ -0,0 +1,319 @@
...
+use crate::util::{Fallible, TryCollectExt};
+use std::error::Error;
+use std::fmt::Write;
+use std::fs::File;
+use std::path::{Path, PathBuf};
+
+#[allow(clippy::nonminimal_bool)]
+pub fn parse(input: &str, path: &Path, settings: &ParserSettings) -> Fallible<Document> {
+    #[derive(Default)]
+    struct State {
```
adamreichold, line 17: I don' think this type really adds anything. Why not go for `let mut state = None;` instead?

src/parse.rs
```diff
@@ -0,0 +1,319 @@
...
+        fn error(err: Box<dyn Error>, line: usize) -> Self {
+            Self::Error(format!("{} (line {})", err, line).into())
+        }
+    }
+
+    let mut state = State::default();
+    let mut document = input
+        .lines()
+        .enumerate()
+        .scan(&mut state, |state, (line_number, line)| {
```
adamreichold, line 37: While this looks nicely functional, I am not sure if implementing the parser via `scan` and `try_collect` is really helpful. I would suggest at least giving a plain for over the numbered lines that can manually aggregate the nodes and errors a try. (Especially as such a loop could just access `nodes.last_mut()` instead of `take()`ing and set `state.node`.)

On a related note, I suspect it would be helpful to factor out parts of this parser into separate functions as well afterwards.

src/parse.rs
```diff
@@ -0,0 +1,319 @@
...
+        if let Some(index) = rest.find(&settings.transclusion_end) {
+            let trans = &rest[..index];
+
+            let target = LINK_REGEX
+                .captures_iter(trans)
+                .map(|match_| match_.get(2).unwrap().as_str())
+                .next()
+                .unwrap_or(&trans);
+
+            let mut path = PathBuf::from(into.parent().unwrap_or_else(|| Path::new(".")));
```
adamreichold, line 209: I am not sure, but maybe this could be simplified as
```
let path = into.parent().unwrap_or_else(|| Path::new(".")).join(target);
```
?

src/parse.rs
```diff
@@ -0,0 +1,319 @@
...
+        Ok(None)
+    }
+}
+
+/// Parses a line as code, returning the parsed `Line` object
+fn parse_line(line_number: usize, input: &str, settings: &ParserSettings) -> Fallible<Line> {
+    let indent_len = input
+        .chars()
+        .take_while(|ch| ch.is_whitespace())
+        .collect::<String>()
```
adamreichold, line 229: Why not just `.count()`?

src/parse.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    Ok(Line {
+        line_number,
+        indent: indent.to_owned(),
+        source: Source::Source(rest.to_owned()),
+        comment,
+    })
+}
+
+pub fn find_links(
```
adamreichold, line 264: This seems to conflate two things: Finding (and returning) the links in the document and removing the markers which would probably best happen already during parsing or at least as a separate post-processing step?

src/print.rs
```diff
@@ -0,0 +1,318 @@
...
+use std::collections::HashMap;
+use std::fmt::Write;
+
+/// Formats this `Document` as a string containing the documentation file contents
+pub fn print_docs(document: &Document, settings: &ParserSettings) -> String {
+    let mut output = String::new();
+    for node in &document.nodes {
+        match node {
+            Node::Transclusion(transclusion) => {
+                output.push_str(&print_transclusion(transclusion, false))
```
adamreichold, line 15: To avoid all these intermediate allocations, it would probably be best if all `print_*` functions would take a `fmt: &mut fmt::Formatter` argument instead of producing a `String`. This would ensure that they combine naturally but producing a `String` as the result would still be simple.

src/print.rs
```diff
@@ -0,0 +1,318 @@
...
+            }
+            Node::Text(text_block) => {
+                output.push_str(&text_block.to_string());
```
adamreichold, line 18: Why not just
```
writeln!(output, "{}", text_block).unwrap();
```
?

src/print.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/// Formats this `Document` as a string containing the compiled code
+pub fn print_code(
+    document: &Document,
+    entrypoint: Option<&str>,
+    language: Option<&str>,
+    settings: Option<&LanguageSettings>,
+) -> Result<String, CompileError> {
+    let comment_start = settings
```
adamreichold, line 50: Wouldn't it make sense to factor out
```
let block_labels = settings.and_then(|settings| settings.block_labels.as_ref());
```
?

src/print.rs
```diff
@@ -0,0 +1,318 @@
...
+        .and_then(|s| s.block_labels.as_ref())
+        .map(|l| l.block_end.as_str())
+        .unwrap_or_default();
+    let block_next = settings
+        .and_then(|s| s.block_labels.as_ref())
+        .map(|l| l.block_next.as_str())
+        .unwrap_or_default();
+    let block_name_sep = '#';
+
+    let clean = if let Some(s) = settings {
```
adamreichold, line 72: This probably wants to use `settings.map_or(true, |settings| ...)`?

src/print.rs
```diff
@@ -0,0 +1,318 @@
...
+                    let count = block_count.entry(&block.name).or_default();
+                    *count += 1;
+                    *count - 1
+                };
+
+                let path = block.source_file.to_owned().unwrap_or_default();
+                let name = if block.is_unnamed {
+                    ""
+                } else {
+                    block.name.as_ref().map(|n| &n[..]).unwrap_or("")
```
adamreichold, line 94: `as_deref()`?

src/print.rs
```diff
@@ -0,0 +1,318 @@
...
+                        .collect::<Vec<_>>()
+                        .join("\n"),
+                )
+            }
+        }
+    }
+    output
+}
+
+pub fn print_transclusion(transclusion: &Transclusion, reverse: bool) -> String {
```
adamreichold, line 195: This looks suspiciously like two unrelated functions shoved into one using a `bool` flag.

src/print.rs
mlange-42: You see an advantage here by avoiding the push of a newline a a further line of code?

adamreichold: The main point here would to avoid the temporary string allocation implied by `.to_string()`. But generally speaking, standardising on `fmt::Formatter` makes the temporaries go away and makes the code mode composable.

src/print.rs
mlange-42: What should I then use as argument? To use `String`, it would require `dyn Write` instead of `Formatter`...

src/print.rs
mlange-42: Yes, definitely. But see https://github.com/mlange-42/yarner/pull/89#discussion_r571621756

src/print.rs
adamreichold: Either generics like `&mut impl Write` or as just suggest `&mut dyn Write`. Since we will probably use each function in once specific case, using `&mut impl Write` seems preferable as there is little fear of code bloat and hence no reason to eat the overhead of dynamic dispatch. (Note that `&mut impl Write` is basically shorthand for `&mut F` and `where F: Write`.)

src/parse.rs
mlange-42: I will give it a try. I also think the use of `scan` makes it unnecessary hard to grasp.

src/print.rs
mlange-42: > (Note that `&mut impl Write` is basically shorthand for `&mut F` and `where F: Write`.)

I was already starting to ask myself why not do it this way 😆

src/print.rs
mlange-42: Made intermediate print functions use `Write` instead of returning `String`

src/parse.rs
mlange-42: This is probably best combined with https://github.com/mlange-42/yarner/pull/89#discussion_r571616614 and https://github.com/mlange-42/yarner/pull/89#discussion_r571617366 into a separate MR. See #97.

src/parse.rs
mlange-42: See #97.

src/parse.rs
mlange-42: See #97.

PR #93, opened 2021-02-06
Simple block labels without `..`

### LINKED ISSUE #92: [Bug] Simplify block label paths when containing `..`
E.g. this:

```
// <@experiments\../resources/random.md#file:src/resources/random.rs#0
```

should instead look like this:

```
// <@resources/random.md#file:src/resources/random.rs#0
```

(This kind of paths occurs through transclusions)

mlange-42:
Fixes #92.

### DIFF
--- a/src/document/code.rs
+++ b/src/document/code.rs
@@ -5,6 +5,7 @@ use crate::config::LanguageSettings;
 use crate::util::TryCollectExt;
 use std::collections::HashMap;
 use std::fmt::Write;
+use std::path::PathBuf;
 
 /// A `CodeBlock` is a block of code as defined by the input format.
 #[derive(Clone, Default, Debug)]
@@ -24,7 +25,7 @@ pub struct CodeBlock {
     /// The source is the lines of code
     pub source: Vec<Line>,
     /// Source file, for transcluded blocks
-    pub source_file: Option<String>,
+    pub source_file: Option<PathBuf>,
 }
 
 impl CodeBlock {
@@ -196,7 +197,7 @@ impl Line {
                             &self.indent,
                             comment_start,
                             if idx == 0 { &block_start } else { &block_next },
-                            path,
+                            path.display(),
                             block_name_sep,
                             name,
                             block_name_sep,
@@ -223,7 +224,7 @@ impl Line {
                             &self.indent,
                             comment_start,
                             &block_end,
-                            path,
+                            path.display(),
                             block_name_sep,
                             name,
                             block_name_sep,
--- a/src/document/code.rs
+++ b/src/document/code.rs
@@ -5,7 +5,6 @@ use crate::config::LanguageSettings;
 use crate::util::TryCollectExt;
 use std::collections::HashMap;
 use std::fmt::Write;
-use std::path::PathBuf;
 
 /// A `CodeBlock` is a block of code as defined by the input format.
 #[derive(Clone, Default, Debug)]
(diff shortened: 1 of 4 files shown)


src/document/code.rs
```diff
@@ -196,7 +197,7 @@ impl Line {
                             &self.indent,
                             comment_start,
                             if idx == 0 { &block_start } else { &block_next },
-                            path,
+                            path.display(),
```
adamreichold, line 200: This does have a conceptual issue: While the result of `.display()` is safe to format and display to humans, it is not guaranteed to be lossless, i.e. it should be avoided if we want to parse this path back later.

There are various ways to go about this, but generally speaking, enforcing valid UTF-8 for paths that are "part of the project structure" seems a reasonable approach in this.

src/parser/md.rs
```diff
@@ -129,7 +129,9 @@ Please comment out option `comments_as_aside` until the next version, and rename
                     .unwrap_or(&trans);
 
                 let mut path = PathBuf::from(into.parent().unwrap_or_else(|| Path::new(".")));
-                path.push(target);
+
+                let sub_path = PathBuf::from(target);
```
adamreichold, line 133: Why do we need to allocate a temporary `PathBuf` if we are appending this to another `PathBuf` anyway?

src/util.rs
```diff
@@ -62,6 +62,29 @@ where
 {
 }
 
+pub trait PushSimple<P> {
+    fn push_simple(&mut self, other: P);
+}
+
+impl PushSimple<PathBuf> for PathBuf {
+    fn push_simple(&mut self, other: PathBuf) {
```
adamreichold, line 70: I would suggest leaving out the type parameter and going for
```rust
fn push_simple(&mut self, other: &Path) { ... }
```
as the generality of the type parameter seems unnecessary and `other: PathBuf` forces a temporary allocation as written above.

src/util.rs
```diff
@@ -62,6 +62,29 @@ where
...
+        for comp in other.components() {
+            match comp.as_os_str().to_str() {
```
adamreichold, line 72: Why do you need to enforce valid UTF-8 here? You should be able to check the component like
```
if comp.as_os_str() == OsStr::new("..") { ... }
```

adamreichold:
As a general observation, this seems to implement a subset of the transformations that our `path-clean` dependency does. Why not just use that instead?

mlange-42:
> As a general observation, this seems to implement a subset of the transformations that our path-clean dependency does. Why not just use that instead?

Yes, you are right. Changed to use the crate for path cleaning. However, path_clean works by string splitting and does not handle `\` (as produced by `PathBuf`). Maybe a `Path`-based solution wold be worth a small crate to publish?

adamreichold (approved):
> However, path_clean works by string splitting and does not handle \ (as produced by PathBuf).

This is intentional and part of a philosophical split on how to handle path name look ups, c.f. https://9p.io/sys/doc/lexnames.html, for the rationale behind the approach in `path-clean`.

> Maybe a Path-based solution wold be worth a small crate to publish?

Looking at the implementation, I think it would make sense to reimplement it based on `Path` and propose that as a change to the existing crate first. I'll start looking into that [here](https://github.com/adamreichold/path-clean)...

adamreichold:
I opened https://github.com/danreeves/path-clean/pull/10 but there is also https://github.com/danreeves/path-clean/pull/7 so we should get this via path-clean eventually one way or the other.

PR #94, opened 2021-02-06
Include file path in error message on failed read

mlange-42:


src/main.rs
```diff
@@ -303,7 +303,8 @@ fn reverse(
                     .get(language)
                     .and_then(|lang| lang.block_labels.as_ref())
                 {
-                    let source = fs::read_to_string(&file).map_err(|err| err.to_string())?;
+                    let source = fs::read_to_string(&file)
+                        .map_err(|err| format!("{}: {}", err.to_string(), file.display()))?;
```
adamreichold, line 307: The call to `.to_string()` is unnecessary (but creates a temporary `String`) as you formatting this `err` into a `String` anyway via `format!`, e.g.
```rust
format!("{}: {}", err, file.display())
```

src/main.rs
```diff
@@ -556,7 +558,8 @@ fn transclude_dry_run(
 }
 
 fn transclude(parser: &MdParser, file_name: &Path) -> Fallible<Document> {
-    let source_main = fs::read_to_string(&file_name)?;
+    let source_main = fs::read_to_string(&file_name)
```
adamreichold, line 561: Since this is done multiple times, maybe a function in `util` like
```rust
pub fn read_file(path: &Path) -> Fallible<String> {
  read_to_string(&path).map_err(|err| format!("{}: {}", err, path.display()))
}
```
would be useful?

adamreichold (approved):

PR #98, opened 2021-02-07
Rework parser loop and link extraction

### LINKED ISSUE #97: [Design] Make parsing a simple for loop, parse links there
* https://github.com/mlange-42/yarner/pull/89#discussion_r571616614
* https://github.com/mlange-42/yarner/pull/89#discussion_r571617366
* https://github.com/mlange-42/yarner/pull/89#discussion_r571618504

mlange-42:
Fixes #97

Part of #57

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -514,16 +514,16 @@ fn transclude_dry_run(
     language: Option<&str>,
     documents: &mut HashMap<PathBuf, Document>,
     track_code_files: &mut HashSet<PathBuf>,
-) -> Fallible<Document> {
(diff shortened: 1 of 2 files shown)


mlange-42:
@adamreichold Integrating link extraction into parsing rather then as post-processing of the document (after transclusions are applied) is not so trivial. The path of the "final" (transcluding) document is required. An option would be to pass the path to `parse` (through `transclude`) as a parameter

adamreichold:
@mlange-42 I saw the review request but please don't hold your breath. I will not be able to do this today.

PR #106, opened 2021-02-09
Prevent overwrite using hashes

### LINKED ISSUE #84: [Feature] Ask for confirmation before overwriting changes (forward/reverse)
When there is the risk of overwriting changes, warn and ask for confirmation. E.g.:

* Markdown sources were changed and the user calls `yarner reverse`
* Code output was changed and the user calls `yarner`

~~Could possibly be detected by analyzing time/date of last change of files.~~

We will use a file `Yarner.lock` for this, which contains hashes over Markdown files and code files todetect changes.

## Considerations

* In forward mode, documents are processed one by one. It would be necessary to collect all code putput file before processing, which requires parsing all documents.
  * Possible solutions:
    * Collect and parse all documents, keep them in memory, check for changes in code files, and then write code it ok.
    * Check the entire code output folder. This would then also cover copied files. Assumes that...
      * ... the user does not use any `..` in code block target files (i.e. everything is indeed in the code outpur folder)
      * ... the code output folder is not above the source folder in the hierarchy
* In reverse mode, it would be sufficient to only hash the code blocks of the sources, as the rest is not changed by this mode

mlange-42:
Store hashes of each source as well as code file in `Yarner.lock`. Use hashes to prevent accidential overwrite.

Fixes #84


Cargo.toml
```diff
@@ -22,3 +22,5 @@ glob = "0.3"
 toml = "0.5"
 clap = "2.32"
 once_cell = "1.5"
+md5 = "0.7"
```
adamreichold, line 25: Please consider using a contemporary hash function, e.g. the [`blake3`](https://crates.io/crates/blake3) crate provides a faster hash with better statistical properties. It even supports hex formatting out of the box...

src/lock.rs
```diff
@@ -0,0 +1,96 @@
...
+        hex::encode(self.hasher.compute().as_ref())
+    }
+
+    pub fn consume_file<P: AsRef<Path>>(&mut self, file: P) -> std::io::Result<()> {
+        let f = File::open(file)?;
+        let mut buffer = [0; 1024];
+        let mut reader = BufReader::new(f);
+
+        loop {
+            let count = reader.read(&mut buffer)?;
```
adamreichold, line 87: If you already use a `BufReader`, you should not copy into another buffer but use `reader.fill_buf()` and `reader.consume()` instead.

src/lock.rs
```diff
@@ -0,0 +1,96 @@
...
+impl Default for Hasher {
+    fn default() -> Self {
+        Self {
+            hasher: Context::new(),
+        }
+    }
+}
+
+impl Hasher {
+    pub fn consume_all<P: AsRef<Path>>(&mut self, path: P) -> std::io::Result<()> {
```
adamreichold, line 64: It seems like you hash only file contents but not paths. So if I do not change the content but rename a file, the hash does not change?

Related to this, I suspect the contents of `Yarner.toml` should probably be included for both source and code as the options can influence the output as well?

src/lock.rs
adamreichold: (But actually in this case of many small files, reading the whole file using `fs::read` and hashing this as a single chunk is probably simpler and faster. Memory should not be an issue as you are doing that everywhere else anyway.)

src/lock.rs
mlange-42: Thanks for pointing that out! I will definitely include paths.

For the `Yarner.toml` I am not really sure. Changes here could break...

* Reverse mode, if code labels change
* The document (e.g. name or macro syntax changes)

Also, if we hash that file in both directions, does that not lock everything once I change it? So you need to `--force` after any change...

src/lock.rs
```diff
@@ -0,0 +1,96 @@
...
+        let root = PathBuf::from(path.as_ref());
+        if root.is_dir() {
+            for entry in fs::read_dir(root)? {
```
adamreichold, line 67: There is another issue here: The order in which directories are enumerated is unspecified but the hash depends on the order. Hence you need to e.g. sort the entries by name to get a stable order for the hash.

src/lock.rs
adamreichold: > Also, if we hash that file in both directions, does that not lock everything once I change it?

I was coming from incremental computation, i.e. only recompute something if the inputs have changed. But you want to use this in the opposite direction, do not do something if the output has been changed since the last computation. So not including `Yarner.toml` should be fine because it isn't overwritten in one direction or the other.

adamreichold: (Hashing paths might actually be weird because on `path::Path` cannot be accessed as `&[u8]` on non-Unix platforms. Probably `.as_os_str().to_string_lossy()` or `.to_str().ok_or("I really do not process non-UTF-8 paths my friend.")`...)

adamreichold:
I have more conceptual concern: This seems to hash the whole directory even if Yarner will not overwrite all files present there, i.e. it would consider untracked files as well. Shouldn't the hashing be limited to the files that Yarner actually touches via copying or printing?

src/lock.rs
```diff
@@ -0,0 +1,85 @@
...
+                self.consume_all(entry.path())?;
+            }
+        } else {
+            self.consume_file(root)?;
+        }
+        Ok(())
+    }
+
+    pub fn compute(self) -> String {
+        format!("{}", self.hasher.finalize().to_hex())
```
adamreichold, line 76: Isn't this just `.to_string()` or actually `.to_owned()`?

src/lock.rs
```diff
@@ -0,0 +1,85 @@
...
+
+        Ok(lock.code_hash != code_hash)
+    } else {
+        Ok(false)
+    }
+}
+
+/// Content for Yarner.lock files
+#[derive(Serialize, Deserialize)]
+pub struct Lock {
```
adamreichold, line 28: Does this and its inherent methods need to be `pub`?

src/lock.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    pub fn write<P: AsRef<Path>>(&self, path: P) -> Fallible {
+        let str = toml::to_string(self)?;
+        write(path, str)?;
+
+        Ok(())
+    }
+}
+
+pub struct Hasher {
```
adamreichold, line 49: Does this and its inherent methods need to be `pub`?

src/lock.rs
mlange-42: I think it is not really unspecified, but depends on the platform. In normal cases it shoud thus be sufficient, but sorting is probably safer.

adamreichold: > I think it is not really unspecified, but depends on the platform. In normal cases it shoud thus be sufficient, but sorting is probably safer.

No it definitely neither specified nor stable. You can list the same directory multiple times and it will yield a different order. Due to VFS cache it can be completely unrelated to any on-disk structure.

adamreichold:
> Shouldn't the hashing be limited to the files that Yarner actually touches via copying or printing?

I would say this suggest a slightly different approach: When processing in forward or reverse mode, write all touched paths and their hashes into the lock file. Before running, check whether the hashes of all those paths in the lock file are unchanged. Slightly larger lock file, but also no issues with hashing unrelated directories and files...

src/lock.rs
mlange-42: by moving some code from main to a function in lock I can make all structs there private

src/lock.rs
mlange-42: same as above

mlange-42:
So you mean writing a hash per file?

For the reverse mode, one could go as far as only hashing code blocks in source files, as the text may change without collisions.

mlange-42:
> I have more conceptual concern: This seems to hash the whole directory even if Yarner will not overwrite all files present there, i.e. it would consider untracked files as well. Shouldn't the hashing be limited to the files that Yarner actually touches via copying or printing?

Principally, yes. For the reverse direction, see my above comment.

I was, however, not sure if I should really start with the full-featured, super exact solution. Same here for all files vs. collected during parsing.

mlange-42:
Meanwhile, I realize that listing files with individual hashes makes an otherwise necessary pre-pass (to determine the affected files) unnecessary. 👍

src/lock.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    fn write<P: AsRef<Path>>(&self, path: P) -> Fallible {
+        let str = toml::to_string(self)?;
+        write(path, str)?;
+
+        Ok(())
+    }
+}
+
+struct Hasher {
```
adamreichold, line 59: Since you really just hash the content of a single file, the free function `blake3::hash` should suffice and `Hasher` can go away completely.

src/lock.rs
```diff
@@ -0,0 +1,86 @@
+use std::fs::write;
+use std::path::{Path, PathBuf};
+
+use serde::{Deserialize, Serialize};
+
+use crate::{files, util::Fallible};
+use std::collections::{HashMap, HashSet};
+
+pub fn code_changed<P: AsRef<Path>>(lock_file: P) -> Fallible<bool> {
+    if lock_file.as_ref().exists() && lock_file.as_ref().is_file() {
```
adamreichold, line 10: The second check includes the first. More generally, it should suffice to just call `Lock::read` and handle any IO and parsing errors there.

adamreichold:
> Meanwhile, I realize that listing files with individual hashes makes an otherwise necessary pre-pass (to determine the affected files) unnecessary. +1

Yes, this was my motivation for suggesting it. I think the is one corner case though: If the user changed things so that a file is written that was not written before, it will not be present in the lock file. But that will only become visible during processing when we actually try to overwrite an already present file whose path is not in the lock file. So we would need to raise a late error in that case and keep the lock file contents (or at least the paths) around during processing.

src/lock.rs
adamreichold: (Or put differently, a free function like `hash_file(path: &Path) -> Fallible<String>` should suffice.)

src/lock.rs
adamreichold: This point seems moot if individual paths and hashes are listed...

src/lock.rs
mlange-42: My intention here was to skip the test if no lock file exists, not to prevent errors. The first condition can be removed, but reading could also fail due to a malformed toml file.

adamreichold: > The first condition can be removed, but reading could also fail due to a malformed toml file.

But do you want to handle this differently than a non-existant file?

src/lock.rs
```diff
@@ -0,0 +1,103 @@
...
+    code_hashes: HashMap<PathBuf, String>,
+}
+
+fn ordered_map<S, K, V>(value: &HashMap<K, V>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+    K: Ord + Serialize,
+    V: Serialize,
+{
+    let ordered: BTreeMap<_, _> = value.iter().collect();
```
adamreichold, line 56: Why not put the `BTreeMap` direct inside `struct Lock`?

src/lock.rs
mlange-42: I think a mal-formed file should result in an error. A missing files could mean that 1) this is the first run with reverse-mode enabled, or 2) the user deleted the file to get rid of a lock. I these cases, it shoud just skip silently and create the lock file after processing.

src/lock.rs
mlange-42: But then, files are in order of insertion I think. We should then sort nevertheless...

src/lock.rs
adamreichold: Valid, but then I would suggest to include a suggestion to delete the malformed file (or use the `--force` flag?) in the error message so that users know how to handle a malformed file.

src/lock.rs
```diff
@@ -0,0 +1,91 @@
...
+            Ok(hash) => Ok((p.clone(), hash)),
+            Err(err) => Err(err),
+        })
+        .collect::<Result<HashMap<_, _>, _>>()
+}
+
+fn hash_file<P: AsRef<Path>>(file: P) -> Fallible<String> {
+    let mut hasher = blake3::Hasher::new();
+    let bytes = files::read_file(file.as_ref())?;
+    hasher.update(&bytes);
```
adamreichold, line 53: Why not just `blake3::hash(bytes).to_hex().to_string()`?

src/lock.rs
mlange-42: It seems I still need a serializer function to get "clean" paths.

adamreichold: Couldn't you just do that in the place where you currently do
```
Ok(hash) => Ok((p.clone(), hash)),
```
?

src/lock.rs
```diff
@@ -39,10 +39,19 @@ pub fn write_lock<P: AsRef<Path>>(
...
+where
+    P: AsRef<Path>,
+{
     files
         .map(|p| match hash_file(p) {
-            Ok(hash) => Ok((p.clone(), hash)),
+            Ok(hash) => Ok((
+                p.as_ref()
+                    .to_str()
+                    .unwrap_or("non-uft8-path")
```
adamreichold, line 51: Won't that fail when you read the lock file? Why not raise an error here?

mlange-42: Yes, but the paths could also come from collecting during processing (used in `write_lock`), not only from the lock file. But maybe we should fail here?

PR #120, opened 2021-02-17
Added support for CRLF line endings through option in Yarner.toml

### LINKED ISSUE #119: [Fix] Handle Windows line endings (esp. in reverse mode)
Internally, Yarner works with Unix line endings (LF / `\n`). Thus, all output has LF line endings. In reverse mode, this results in Windows line endings (CRLF / `\r\n`) being converted to Unix line endings.

E.g., Git may then show all source files as changed, but no diffs, or a lot of warnings.

**Possible solutions:**

1. Respect line endings, based on either
    * Configurable in the `Yarner.toml` (see #120), or
    * The user's system (see #120, too), or
    * Existing line endings
1. Recommend the use of Unix line endings, and provide the required Git configuration

mlange-42:
Line endings can be controlled via bool option `crlf_newline` in section `[parser]` of `Yarner.toml`.

~~Not really an optimal solution for projects used on different systems.~~

Made `crlf_newline` optional, use system newline if not provided.

Fixes #119

### DIFF
--- a/guide/src/config-parser.md
+++ b/guide/src/config-parser.md
@@ -26,6 +26,8 @@ link_prefix = "@"
 
 file_prefix = "file:"
 hidden_prefix = "hidden:"
+
+# crlf_newline = false
 ```
 
 ## Options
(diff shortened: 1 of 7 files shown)


mlange-42:
Withdrawn in favor of #122

PR #121, opened 2021-02-18
Detect newline type from documents

mlange-42:


src/parse.rs
```diff
@@ -105,7 +113,20 @@ pub fn parse(
         return Err(msg.into());
     }
 
-    Ok((Document::new(nodes), links))
+    Ok((Document::new(nodes, newline), links))
+}
+
+fn detect_newline(text: &str) -> &'static str {
+    NEWLINE_REGEX.captures_iter(text).next().map_or_else(
```
adamreichold, line 120: Wouldn't it be simpler to just check `text.contains("\r\n")` and assume Unix line endings otherwise?

adamreichold: (The mixed cases are wrong either way, aren't they?)

src/document.rs
```diff
@@ -6,10 +6,14 @@ use std::collections::HashMap;
 use std::fmt::Write;
 use std::path::{Path, PathBuf};
 
+pub(crate) const CRLF_NEWLINE: &str = "\r\n";
```
adamreichold, line 9: As this is binary, i.e. leaf node in the dependency graph, `pub(crate)` and `pub` are equivalent.

src/parse.rs
mlange-42: Yes, mixed is evil! 😈 

I wanted to simply use the first newline detected, as `text.contains("\r\n")` would use win endings as soon as any "wrong" line ending is found in a unix-style file.

I just completed to re-do everything on a separate, clearer branch, so I will move everything to another MR and close this one.

mlange-42:
Withdrawn in favor of #122

src/parse.rs
mlange-42: Changed as suggested, in #122

mlange-42: But see https://github.com/mlange-42/yarner/pull/122#discussion_r578768880

PR #122, opened 2021-02-18
Derive newline from source documents

### LINKED ISSUE #119: [Fix] Handle Windows line endings (esp. in reverse mode)

mlange-42:
Fixes #119

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1,4 +1,4 @@
-use crate::config::{ParserSettings, CRLF_NEWLINE, LF_NEWLINE, LINK_REGEX, NEWLINE_REGEX};
+use crate::config::{ParserSettings, CRLF_NEWLINE, LF_NEWLINE, LINK_REGEX};
 use crate::document::{CodeBlock, Document, Line, Node, Source, TextBlock, Transclusion};
 use crate::util::Fallible;
 use regex::Captures;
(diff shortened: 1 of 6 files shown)


src/parse.rs
```diff
@@ -105,7 +107,15 @@ pub fn parse(
         return Err(msg.into());
     }
 
-    Ok((Document::new(nodes), links))
+    Ok((Document::new(nodes, newline), links))
+}
+
+fn detect_newline(text: &str) -> &'static str {
+    if text.contains(CRLF_NEWLINE) {
```
mlange-42, line 114: Couldn't the old version with captures potentially be more efficient? In a Unix file, the `contains` version would have to search through the entire string, while the captures version could stop at the first newline, no matter which type it is.

adamreichold: > In a Unix file, the contains version would have to search through the entire string, while the captures version could stop at the first newline, no matter which type it is.

That is the true, but you have already read the whole file into memory so it probably does not matter compared to the cost of a regular expression.

Also, you can still have that behaviour if you really want: Search for the first `'\n'` using `.find(...)` and then check if it is preceded by a `'\r'`. This will use `memchr` and should highly efficient. But it also does have different semantics in the mixed case.

(Personally, I am not preferring one over the other. I just think that both options do not require the machinery of regular expressions.)

mlange-42: Ok, thanks for the clarification!

mlange-42: Changed to the `find`+`ends_with`-version

adamreichold: This is approaching code golf, but you could avoid the redundant else clauses via early return, e.g.
```rust
if let Some(pos) = text.find('\n') {
  if text[..pos].ends_with('\r') {
    return CRLF_NEWLINE;
  }
}

LF_NEWLINE
```

mlange-42: Like the shoter code! :wink:

PR #125, opened 2021-02-20
JSON I/O to let pre-processors modify documents

mlange-42:
* Move JSON-relevant modules to library crate
* Add pre-processor section to config
* run all pre-processors in the order given in config


mlange-42:
Implemented more incrementally in #127, #128, #129 and #130

PR #160, opened 2021-03-05
Remove option `arguments`, parse command string using shlex

mlange-42:


guide/src/plugins/using.md
```diff
@@ -35,10 +35,9 @@ Plugins have no effect in reverse mode.
...
 By default, the command of the plugin is derived from its name in the config file, prefixed with `yarner-`. E.g., the command derived from `[plugin.block-links]` is `yarner-block-links`.
 
-Alternatively, each plugin section can have optional parameters `command` and `arguments`, e.g.
+Alternatively, each plugin section can have optional parameters `command`, e.g.
 
 ```toml
 [plugin.xyz]
-command = "path/to/binary"
-arguments = ["--arg1", "value1", "--arg2", "value2"]
+command = "path/to/binary --arg value"
```
adamreichold, line 42: I think introducing the complexities of shell argument splitting even though a usable list syntax is available via TOML is a decrease in usability. These argument splitting procedures only exists because the early shells lacked a proper list syntax.

Consider for example using a  argument value containing a space
```toml
command = "path/to/binary"
arguments = ["--arg", "value with space"]
```
and
```toml
command = "path/to/binary --arg \"value with space\""
```
or something similar using single quotes.

This has the additional usability surprise that it might look like this is passed through a shell, but it is actually not, only the arguments are split according to the systems shell syntax, e.g.
```toml
command = "path/to/binary --arg $HOME/value"
```
will not work. (Put differently, If the complexities of shell syntax are taken up, then it would seem preferable to actually pass this through a shell instead of parsing it using `shlex`.)

mlange-42: Thanks a lof for these considerations! The usability surprise is indeed a striking argument. I will consider reverting this.  
I would still be interested whether passing it through a shell is possible for the platforms supported, and how passing the JSON would then look like.

adamreichold: > I would still be interested whether passing it through a shell is possible for the platforms supported

It certainly is, e.g. on Unix by running
```rust
Command::new("/bin/sh").arg("-c").arg(command)
```
and connecting stdin and stdout as usual. (It is the stdin/out of the shell process but that is inherited to the applications started by the shell.) The rest of the interface should not change at all.

adamreichold: (I am not sure whether a good crate to do this portably exists, but it should as call [`libc::system`](https://en.cppreference.com/w/c/program/system) should work portably as a last resort but without the amenities of `std::process::Command`.)

adamreichold: Thinking about this a bit more, I would really advise against calling a shell here: This requires users to use two different syntaxes (TOML and shell) to configure Yarner and it does not really buy them anything w.r.t. flexibility in return. Those who want scripting using shell or Python etc. can just create an executable script which can then be exec'ed like any other binary.

mlange-42: While currently browsing a bit for a rust crate for shell commands, I start to come to the same conclusion.

mlange-42: Reverting in #161

PR #161, opened 2021-03-06
Revert to use option `arguments`

mlange-42:
See https://github.com/mlange-42/yarner/pull/160#discussion_r588819953


src/plugin.rs
```diff
@@ -20,7 +20,21 @@ pub fn run_plugins(
             .and_then(|cmd| cmd.as_str().map(|s| s.to_owned()))
             .unwrap_or_else(|| format!("yarner-{}", name));
 
-        let args = shlex::split(&command).ok_or(format!("Invalid plugin command {}", command))?;
+        let arguments: Vec<&str> = match config.get("arguments") {
+            None => vec![],
```
adamreichold, line 24: Isn't there a `.map(..).unwrap_or_default()` in here somewhere as well?

mlange-42: Do you mean replacing the match?

src/plugin.rs
```diff
@@ -20,7 +20,21 @@ pub fn run_plugins(
...
+            Some(v) => v
+                .as_array()
+                .map(|arr| arr.iter().map(|l| l.as_str().unwrap_or_default()))
+                .ok_or("Can't parse array of plugin arguments")?
+                .collect(),
+        };
+
+        let command_string = format!(
+            "{}{}{}",
+            command,
```
adamreichold, line 34: Couldn't you just `iter::once(command).chain(arguments).join(" ")`?

src/plugin.rs
adamreichold: Yes, as it basically maps `None` to `Vec`'s `Default` impl.

src/plugin.rs
mlange-42: Didn't know of this "syntax"...

Although it is a little more complicated...

```rust
let command_string = std::iter::once(&command.as_str())
    .chain(&arguments)
    .join(" ", "");
```

adamreichold: It is a matter of preference I guess. One might even drop printing the arguments entirely as the printed command might also be misleading without quoting the individual arguments (spaces and such).

src/plugin.rs
mlange-42: Due to a `Option<Result<_, _>>`, I only get it to work like this:

```rust
let arguments: Vec<&str> = config
            .get("arguments")
            .map(|args| {
                args.as_array()
                    .map(|arr| arr.iter().map(|l| l.as_str().unwrap_or_default()).collect())
                    .ok_or("Can't parse array of plugin arguments")
            })
            .unwrap_or_else(|| Ok(vec![]))?;
```

src/plugin.rs
mlange-42: I think the chain version is better than the previous one, particularly due to the condition inside `format!`. Your second point is definitely worth considering.

mlange-42: Simplified plugin log messages and removed this.

src/plugin.rs
adamreichold: I did not compile this, but there is [`transpose`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.transpose) to go from `Option<Result<T, E>>` to `Result<Option<T>, E>`, e.g.
```rust
let arguments: Vec<&str> = config
            .get("arguments")
            .map(|args| {
                args.as_array()
                    .map(|arr| arr.iter().map(|l| l.as_str().unwrap_or_default()).collect())
                    .ok_or("Can't parse array of plugin arguments")
            })
            .transpose()?
            .unwrap_or_default();
```

PR #162, opened 2021-03-08
Watch command

mlange-42:
Successfully implemented detection of file changes in forward (sources changed) and reverse (code changed) direction.

Now there is a problem that could have been forseeable: Say a change in the sources is detected. Yarner builds and produces (i.e. modifies) the code. Code files have changed, and a reverse build starts. Source files change, etc...

Moved code out of main, which makes the changes seem more than they are.


src/watch.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub fn watch(
+    args: ArgMatches,
+    watch_sources: HashSet<PathBuf>,
+    watch_code: HashSet<PathBuf>,
+) -> Fallible {
+    trigger_on_change(&watch_sources, &watch_code, |paths| {
+        let is_forward = paths
+            .first()
+            .map(|p| watch_sources.contains(p))
```
mlange-42, line 17: Not really clear to me how to distinguish the two watchers. Is there any better way to detect where the event came from?

adamreichold: Use two distinct channels for both watchers. As selection is not possible using `std`'s channels anymore, spawn separate threads to merge results using an `enum`, e.g.
```rust
enum ChangeType {
  Sources,
  Code
}

spawn(move || {
  for event in rx_sources {
    tx_changes.send(ChangeType::Source).unwrap();
  }
});


spawn(move || {
  for event in rx_code {
    tx_changes.send(ChangeType::Code).unwrap();
  }
});
```

Would have been simpler if they'd used `crossbeam-channel` which does support `select!`, but then again, `notify` spawns a thread per watcher anyway.

src/watch.rs
```diff
@@ -0,0 +1,77 @@
...
+    for path in watch_sources {
+        source_watcher.watch(path, NonRecursive)?;
+    }
+    for path in watch_code {
+        code_watcher.watch(path, NonRecursive)?;
+    }
+
+    loop {
+        let first_event = rx.recv().unwrap();
+        sleep(Duration::from_millis(50));
```
adamreichold, line 61: This does not look right. I think you want to coalesce all events within 50ms into a single change? How about
```rust
rx.recv().unwrap();
let deadline = Instant::now() + Duration::from_millis(50);

loop {
  let timeout = match deadline.checked_sub(Instant::now()) {
    Some(timeout) => timeout,
    None => break,
  };
  
  rx.recv_timeout(timeout);
}

tx_changes.send(ChangeType::Sources).unwrap();
```
which could be done directly in the post-processing threads suggested above.

src/watch.rs
mlange-42: I do not really understand how these threads relate to the loop we currently have.

mlange-42: Ok, I think I got it...

mlange-42: Not sure if I got that correctly, combined with https://github.com/mlange-42/yarner/pull/162#discussion_r589231253, but at least it works.

mlange-42:
Maybe with the new structuring into two different channels it is easier to solve, but the problem of infinite forth-and-back compilation still exists.

Somehow, events from one channel need to be ignored for a certain time span after an event from the other channel. Or is there a better solution?

adamreichold:
> Somehow, events from one channel need to be ignored for a certain time span after an event from the other channel. Or is there a better solution?

Could you check the log if anything actually changed before reacting to a change event?

Or you could go even further and only write files that have actually changed, i.e. before writing to a file, read it and check if its content is already the same as what you are going to write. I use that approach for the bee model to keep incremental compilation fast which would always recompile apparently modified extracted code.

src/watch.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(PartialEq)]
+pub enum ChangeType {
+    Sources,
+    Code,
+}
+
+pub fn watch(
+    args: ArgMatches,
+    watch_sources: HashSet<PathBuf>,
```
adamreichold, line 18: Do you really need to collect this into a hash table?

src/watch.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use clap::ArgMatches;
+use notify::{DebouncedEvent, Watcher};
+use std::time::Instant;
+use std::{collections::HashSet, env, path::PathBuf, sync::mpsc::channel, time::Duration};
+
+const COLLECT_EVENTS_MILLIS: u64 = 500;
+
+#[derive(PartialEq)]
+pub enum ChangeType {
```
adamreichold, line 11: Does anything here but `watch` really need to be public?

src/watch.rs
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+        if send_event {
+            tx_changes_code.send(ChangeType::Code).unwrap();
+        }
+    });
+
+    loop {
+        let event = rx_changes.recv().unwrap();
+        closure(event)?;
```
adamreichold, line 111: The closure API seems overkill at this point. Why not just return `ex_changes`?

src/watch.rs
```diff
@@ -0,0 +1,123 @@
...
+        code_watcher.watch(path, NonRecursive)?;
+    }
+
+    let (tx_changes_sources, rx_changes) = channel();
+    let tx_changes_code = tx_changes_sources.clone();
+    std::thread::spawn(move || loop {
+        let deadline = Instant::now() + Duration::from_millis(COLLECT_EVENTS_MILLIS);
+        let mut send_event = false;
+        loop {
+            let timeout = deadline.saturating_duration_since(Instant::now());
```
adamreichold, line 73: Why not `checked_duration_since`?

src/watch.rs
```diff
@@ -0,0 +1,123 @@
...
+        source_watcher.watch(path, NonRecursive)?;
+    }
+    for path in watch_code {
+        code_watcher.watch(path, NonRecursive)?;
+    }
+
+    let (tx_changes_sources, rx_changes) = channel();
+    let tx_changes_code = tx_changes_sources.clone();
+    std::thread::spawn(move || loop {
+        let deadline = Instant::now() + Duration::from_millis(COLLECT_EVENTS_MILLIS);
```
adamreichold, line 70: Why do start the deadline before receiving the first event?

src/watch.rs
```diff
@@ -0,0 +1,123 @@
...
+                }
+            }
+        }
+        if send_event {
+            tx_changes_sources.send(ChangeType::Sources).unwrap();
+        }
+    });
+
+    std::thread::spawn(move || loop {
+        let deadline = Instant::now() + Duration::from_millis(COLLECT_EVENTS_MILLIS);
```
adamreichold, line 90: This is a bit redundant as the only difference to the above is why `ChangeType` variant is sent. So it should be straight-forward to pull this out into a function.

src/watch.rs
mlange-42: No, not really. So should I just use `Iterator` (or `&mut Iterator` in `trigger_on_change`)?

src/watch.rs
mlange-42: I think I don't understand that. What is the sense of returning from the loop?

adamreichold: I mean that instead of passing a closure in and calling in that in a loop, do not include the loop in the function at all. Just return the receiver and do
```rust
for event in rx_changes.iter() { ... }
```
on the outside.

src/watch.rs
adamreichold: `impl Iterator<Item=PathBuf>` or something like it should work.

src/watch.rs
mlange-42: I tried to return `rx_changes` from `trigger_on_change`, and use a loop as above in `watch`. However, the cannel then seemy colsed and I get a panic at `watch.rs`, L89 (`let event = in_channel.recv().unwrap();`).

mlange-42:
> Could you check the log if anything actually changed before reacting to a change event?

Yes, this may help but not fix the problem completely. It would then always do two passes (e.g. forward -> code changes -> reverse -> sources do not change -> stop).

> Or you could go even further and only write files that have actually changed, i.e. before writing to a file, read it and check if its content is already the same as what you are going to write. I use that approach for the bee model to keep incremental compilation fast which would always recompile apparently modified extracted code.

Combined with the above, this would probably fix the problem. I think what would happen is: forward -> code changes -> reverse -> does not write due to no changes -> stop.

src/watch.rs
adamreichold: Yes, because the watchers are kept on stack as it is. You can move those onto the threads that handle them using a binding like
```rust
thread::spawn(move || {
    let _watcher = watcher;
...
```

adamreichold:
> Combined with the above,

Do you really need to combine those? Wouldn't no touch unchanged files suffice?

mlange-42:
> Do you really need to combine those? Wouldn't no touch unchanged files suffice?

You are right. With only the second approach, I get the predicted behaviour.

src/compile/forward.rs
```diff
@@ -118,10 +123,15 @@ fn extract_code(
...
                 )?;
-                println!("  Writing file {}", file_path.display());
-                fs::create_dir_all(file_path.parent().unwrap())?;
-                let mut code_file = File::create(file_path)?;
-                write!(code_file, "{}", code)?;
+
+                if files::file_differs(&file_path, &code)? {
+                    println!("  Writing file {}", file_path.display());
+                    fs::create_dir_all(file_path.parent().unwrap())?;
+                    let mut code_file = File::create(file_path)?;
```
adamreichold, line 130: This looks a lot like `fs::write`?

src/compile/forward.rs
```diff
@@ -138,20 +148,29 @@ fn extract_code(
...
             let mut file_path = doc_dir.to_owned();
             file_path.push(file_name);
-            fs::create_dir_all(file_path.parent().unwrap()).unwrap();
-            let mut doc_file = File::create(file_path).unwrap();
-            write!(doc_file, "{}", documentation).unwrap();
+
+            if files::file_differs(&file_path, &documentation)? {
+                println!("Writing documentation file {}", file_name.display());
+                fs::create_dir_all(file_path.parent().unwrap()).unwrap();
+                let mut doc_file = File::create(file_path).unwrap();
```
adamreichold, line 161: As above, this probably wants to be `fs::write`.

src/files.rs
```diff
@@ -19,16 +19,32 @@ pub fn read_file(path: &Path) -> Fallible<Vec<u8>> {
...
+fn files_differ(old: &Path, new: &Path) -> Fallible<bool> {
+    if !(old.is_file() && new.is_file()) {
+        return Ok(true);
+    }
+
+    Ok(read_file(old)? != read_file(new)?)
+}
+
+pub fn file_differs(file: &Path, content: &str) -> Fallible<bool> {
+    if !file.is_file() {
```
adamreichold, line 31: This looks redundant as reading the file would fail otherwise. Why not just turn failure-to-read into a change?

src/files.rs
```diff
@@ -19,16 +19,32 @@ pub fn read_file(path: &Path) -> Fallible<Vec<u8>> {
     std::fs::read(&path).map_err(|err| format!("{}: {}", err, path.display()).into())
 }
 
+fn files_differ(old: &Path, new: &Path) -> Fallible<bool> {
+    if !(old.is_file() && new.is_file()) {
```
adamreichold, line 23: As below, turning I/O errors into changes should suffice.

guide/src/watch-command.md
```diff
@@ -0,0 +1,21 @@
+# Watch command
+
+Yarner has a subcommand `watch` to automatically build a project after files changed. This may be particularly convenient when switching forth and back between Markdown editing and code editing with reverse mode.
+
+<div style="border: 2px solid red; padding: 0.5em;">
+Warning: This feature is still experimental and modifies the original Markdown sources. Make a backup of the sources before using it!
```
adamreichold, line 6: If I understand things correctly, it will also currently not watch `Yarner.toml` and hence not react to changes to the set of source files?

src/compile/forward.rs
mlange-42: Cool, didn't know that... There is also a third occurrence...

src/files.rs
mlange-42: You may have had a less verbose implementation in mind?

guide/src/watch-command.md
mlange-42: It does watch `Yarner.toml`, but only as source file. I.e. it will attempt a forward pass on changes.

A limitation is still that changes in the list of files to watch are not taken into account (e.g. if a transclusion is added).

src/files.rs
```diff
@@ -19,16 +18,37 @@ pub fn read_file(path: &Path) -> Fallible<Vec<u8>> {
...
+    };
+    let new_content = match read_file(new) {
+        Ok(c) => c,
+        Err(_) => return true,
+    };
+    old_content != new_content
+}
+
+pub fn file_differs(file: &Path, content: &str) -> bool {
+    let other_content = match read_file_string(file) {
```
adamreichold, line 34: I think this would be a good candidate for `map_or`, e.g.
```rust
read_file_string(file).map_or(true, |content| content != new_content)
```
and similarly above
```rust
read_file(old).and_then(|old|  read_file(new).map(|new| old != new)).unwrap_or(true)
```

guide/src/watch-command.md
mlange-42: When returning to use a closure, it might be possible to modify the watchers? Will give it a try...

adamreichold: > A limitation is still that changes in the list of files to watch are not taken into account (e.g. if a transclusion is added).

I don't think implementing this in this PR here would be helpful. Just, that it could be included in the documentation.

PR #163, opened 2021-03-08
Update watch list

mlange-42:
* Update the watch list if files changed
* Fix missing tracking/lock of transcluded files

### DIFF
--- a/src/watch.rs
+++ b/src/watch.rs
@@ -51,28 +51,34 @@ where
 
                 watch_sources_new.insert(config);
 
-                for path in &watch_sources_old {
-                    sw.unwatch(path)?;
-                }
-                for path in &watch_code_old {
-                    cw.unwatch(path)?;
-                }
+                update_watcher(sw, &watch_sources_old, &watch_sources_new)?;
+                update_watcher(cw, &watch_code_old, &watch_code_new)?;
 
-                watch_sources_old.clear();
-                watch_code_old.clear();
-                for path in watch_sources_new {
-                    sw.watch(&path, RecursiveMode::NonRecursive)?;
-                    watch_sources_old.insert(path);
-                }
-                for path in watch_code_new {
-                    cw.watch(&path, RecursiveMode::NonRecursive)?;
-                    watch_code_old.insert(path);
-                }
+                watch_sources_old = watch_sources_new;
+                watch_code_old = watch_code_new;
             }
             Ok(())
         },
     )?;
 
+    Ok(())
+}
+
+fn update_watcher(
+    watcher: &mut RecommendedWatcher,
+    old_files: &HashSet<PathBuf>,
+    new_files: &HashSet<PathBuf>,
+) -> Fallible {
+    for path in old_files {
+        if !new_files.contains(path) {
+            watcher.unwatch(path)?;
+        }
+    }
+    for path in new_files {
+        if !old_files.contains(path) {
+            watcher.watch(&path, RecursiveMode::NonRecursive)?;
+        }
+    }
     Ok(())
 }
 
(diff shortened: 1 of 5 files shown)


src/watch.rs
```diff
@@ -26,98 +27,116 @@ where
...
 
-    for change in rx_changes {
-        if allow_reverse || change == ChangeType::Sources {
-            println!(
-                "{} changed. Re-building...",
-                if change == ChangeType::Sources {
-                    "Sources"
-                } else {
-                    "Code"
+                for path in &watch_sources_old {
```
adamreichold, line 54: I think if you go through the trouble of updating the existing watchers (instead of just creating a completely new set of watchers/threads), you should extend this to only remove old paths which are not also part of the new paths and to add only those new paths which where not already part of the old paths.

mlange-42: This can indeed be simplified further...

src/watch.rs
```diff
@@ -26,98 +27,122 @@ where
...
-            env::set_current_dir(&curr_dir)?;
+    Ok(())
+}
+
+fn update_watcher(
+    watcher: &mut RecommendedWatcher,
+    old_files: &HashSet<PathBuf>,
+    new_files: &HashSet<PathBuf>,
+) -> Fallible {
+    for path in old_files {
```
adamreichold, line 72: The [`HashSet::difference`](https://doc.rust-lang.org/stable/std/collections/struct.HashSet.html#method.difference) method is not applicable here?

src/watch.rs
```diff
@@ -26,98 +27,122 @@ where
...
     println!("Watching for changes...");
 
-    let rx_changes = trigger_on_change(&mut watch_sources, &mut watch_code)?;
+    let mut watch_sources_old: HashSet<_> = watch_sources.collect();
+    let mut watch_code_old: HashSet<_> = watch_code.collect();
+
+    trigger_on_change(
+        watch_sources_old.clone(),
+        watch_code_old.clone(),
+        |change, sw, cw| {
```
adamreichold, line 36: One thing I just noticed: You neither need the closure API nor to keep the watchers on the stacks of the background threads, you could just return `(change, sw, cw)` from `trigger_on_change`.

src/watch.rs
mlange-42: Another thing I was not aware of...

src/watch.rs
mlange-42: with `change`, you mean the receiver?

adamreichold: Yes, just returning a tuple of the tree things you pass arguments to the closure.

src/watch.rs
```diff
@@ -40,84 +45,95 @@ where
...
+    }
+    for path in new_files.difference(old_files) {
+        watcher.watch(&path, RecursiveMode::NonRecursive)?;
+    }
+    Ok(())
+}
 
+/// Calls the closure when a book source file is changed, blocking indefinitely.
+fn trigger_on_change(
+    watch_sources: HashSet<PathBuf>,
```
adamreichold, line 81: One consequence of returning instead of the closure should that you do not need ownership here anymore, i.e. `&HashSet<PathBuf`> or even `impl Iterator<Item=PathBuf>` should now suffice. (And hence the cloning be unnecessary.)

src/watch.rs
```diff
@@ -40,84 +45,95 @@ where
...
-        rx_sources,
-        tx_changes.clone(),
-        ChangeType::Sources,
-    );
-    start_event_thread(code_watcher, rx_code, tx_changes, ChangeType::Code);
+    start_event_thread(rx_sources, tx_changes.clone(), ChangeType::Sources);
+    start_event_thread(rx_code, tx_changes, ChangeType::Code);
 
-    Ok(rx_changes)
+    /*loop {
```
adamreichold, line 101: Left-over commented out code?

PR #164, opened 2021-03-08
Suspend "opposing" watcher during build process

mlange-42:
When saving multiple times shortly after another, it may happen that the attempted reverse operation fails due to "Source/code files changed. Stopping to avoid overwriting..."

Thus, simply checking for changes before writing is not sufficien. Suspending the "opposing" watcher also completely avoids the unneccessary attempt for the opposing build direction.

### DIFF
--- a/src/watch.rs
+++ b/src/watch.rs
@@ -8,7 +8,7 @@ use std::sync::mpsc::{Receiver, Sender};
 use std::time::Instant;
 use std::{env, path::PathBuf, sync::mpsc::channel, time::Duration};
 
-const COLLECT_EVENTS_MILLIS: u64 = 500;
+const COLLECT_EVENTS_MILLIS: u64 = 100;
 
 #[derive(PartialEq, Clone)]
 enum ChangeType {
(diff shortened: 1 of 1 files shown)


src/watch.rs
```diff
@@ -73,6 +86,20 @@ fn update_watcher(
     Ok(())
 }
 
+fn unwatch_all(watcher: &mut RecommendedWatcher, old_files: &HashSet<PathBuf>) -> Fallible {
+    for path in old_files {
+        watcher.unwatch(&path)?;
```
adamreichold, line 91: This is probably a rather costly operation as it involves multiple system calls. A more lightweight approach would probably be to share an `suspend: Arc<AtomicBool>` with each background thread which leads to all events being ignored when `suspend.load(Ordering::SeqCst)` is true.

mlange-42: I will give it a try, but not sure if this works with the *debounce* watcher that delays by 1 second.

mlange-42: With an implementation as in #165, it seems not to work. Maybe I missed something, but I think it is due to the delay in debounce watcher.

adamreichold: (As an aside, do you really need the debounced watcher as you already coalesce events anyway? You might want to wait for quiescence though, i.e. reset the timeout whenever an event arrives as there is little use to rebuilding while the files still change.)

mlange-42:
Closed in favor of #165

PR #165, opened 2021-03-08
Suspend watchers using Arc<AtomicBool> and RawWatcher

mlange-42:


src/watch.rs
```diff
@@ -122,7 +140,7 @@ fn start_event_thread(
             };
 
             if let Ok(event) = in_channel.recv_timeout(timeout) {
-                if is_file_change(event) {
+                if is_file_change(event) && !suspend.load(Ordering::SeqCst) {
```
adamreichold, line 143: A proper state machine that suspends changes for a given duration might be necessary, but maybe just doing
```rust
if suspend.load(Ordering::SeqCst) {
  sleep(COLLECT_EVENTS_MILLIS)
} else if send_event {
  out_channel.send(...)
}
```
would work as well. (Possibly together with breaking the inner loop if suspend is set.)

adamreichold: (Of course, all of this is a bit racy as this might suspend reacting to changes for too long. But that applies to the unwatch solution as well but without any control over how long it takes to re-establish the watches.)

mlange-42: Again, I might be missing something, but this does not prevent the "opposite" step to run. The delay I mean is not `COLLECT_EVENTS_MILLIS`, but the one second passed to

```rust
let mut source_watcher = notify::watcher(tx_sources, Duration::from_secs(1))?;
```

So what I think what happens:
1. Switch suspend on
1. Process files, output code
1. Watcher is notified, but delays sending
1. Switch suspend off
1. ... 1sec
1. Watcher sends event

So, as you write, I somehow need to keep suspension for more than the one second of delay.

adamreichold: >  So, as you write, I somehow need to keep suspension for more than the one second of delay.

So the only change is using `sleep(Duration::from_secs(2))` in the suspend branch?

adamreichold: Ah, sleeping alone won't work as the channel buffers as well. So instead of sleeping, you need to drain the channel for two seconds, e.g.
* deadline = now + 2s
* loop:
* timeout = deadline-now or break
* recv_timeout(timeout)

adamreichold: Also as written in the other PR, I don't think you debouncing as you coalesce already, but I think you should use a constant timeout and really wait for quiescence.

mlange-42: So, as done in the last commit? It works, and when saving a file multiple times in high frequency, it waits until the last save.

src/watch.rs
```diff
@@ -41,6 +48,8 @@ pub fn watch(
                 }
             );
 
+            suspend.swap(true, Ordering::SeqCst);
```
adamreichold, line 51: You are not using the previous value, so `suspend.store(true, Ordering::SeqCst);` should suffice here and below.

src/watch.rs
```diff
@@ -95,36 +107,43 @@ where
...
-        if is_file_change(event) {
+        in_channel.recv().unwrap();
+        if !suspend.load(Ordering::SeqCst) {
             send_event = true;
         }
 
-        let deadline = Instant::now() + Duration::from_millis(COLLECT_EVENTS_MILLIS);
+        let mut deadline = Instant::now() + Duration::from_millis(COLLECT_EVENTS_MILLIS);
         loop {
             let timeout = match deadline.checked_duration_since(Instant::now()) {
```
adamreichold, line 137: I think you can drop all this, you just need
```rust
while in_channel.recv_timeout(COLLECT_EVENTS).is_ok() {}
```
now. (It should be possible to have a `const COLLECT_EVENTS: Duration = Duration::from_millis(500)` as `from_millis` is a `const fn` by now.)

src/watch.rs
```diff
@@ -95,36 +107,43 @@ where
...
+    start_event_thread(rx_code, tx_changes, ChangeType::Code, suspend);
 
     Ok((rx_changes, source_watcher, code_watcher))
 }
 
 fn start_event_thread(
-    in_channel: Receiver<DebouncedEvent>,
+    in_channel: Receiver<RawEvent>,
     out_channel: Sender<ChangeType>,
     event_type: ChangeType,
```
adamreichold, line 124: It is not part of the diff here, but I noticed that you need to `.clone()` this below. Why not `derive(Copy)` on the `enum`?

src/watch.rs
adamreichold: Since you are not filtering either anymore, I guess this can be simplified further still, e.g.
```
loop {
  in_channel.recv().unwrap();
  if suspend.load(Ordering::SeqCst) { continue; }

  while in_channel.recv_timeout(COLLECT_EVENTS).is_ok() {}
  if suspend.load(Ordering::SeqCst) { continue; }

  out_channel.send(event_type).unwrap();
}
```

mlange-42: 🎉

PR #166, opened 2021-03-09
Watch mode - less errors, more warnings

mlange-42:
Make using watch more convenient by not aborting on "recoverable" errors

* Warn instead of abort on plugin i/o problem (allow for plugins not modifying the docs)
* Warn instead of abort on plugin returning with error, in watch mode (avoid exiting watch mode)
* Warn instead of abort on unconfigured reverse mode, in watch mode (allow to enable reverse mode while watching)

### DIFF
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -18,6 +18,7 @@ use crate::{
 pub fn run_with_args(
     matches: &ArgMatches,
     reverse_mode: Option<bool>,
+    strict: bool,
 ) -> Fallible<(PathBuf, HashSet<PathBuf>, HashSet<PathBuf>, bool)> {
     let config_path = matches.value_of("config").unwrap();
     let mut config = Config::read(config_path)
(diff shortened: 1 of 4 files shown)


adamreichold:
Long-term, this might be a point where you want to add a structure error type and do the matching on the kind of error on the outside. But then again, I could also understand ignoring any error (justing printing it and continue to watch) in watch mode instead of investing into these fine-grained distinctions.

mlange-42:
In the medium to long term, this sounds like a useful thing. Until then, there will probably show up other kinds of errors that may require different handling in watch mode.

adamreichold:
> other kinds of errors that may require different handling in watch mode.

Do you think there is any need to handle errors (besides printing) in watch mode at all? If you consider using an external watching application, it will normally not fail if the triggered subprocess fails but just retry after further changes as well. If think this would be a simple and use solution: Instead of bubbling up the error in the watch loop, just print it and continue.

PR #177, opened 2021-05-08
Make the JoinExt implementation lazy to avoid the intermediate String allocation

adamreichold:
Taking a hint from the [`display_utils`](https://crates.io/crates/display_utils) crate, this makes the `JoinExt` lazy so that no intermediate must be allocated if the final target is formatting the result (which it usually is).

Thinking further, since the purpose of this crate is pretty much shuffling text around, it might even be more reasonable to add a dependency on `display_utils` replace our bespoke `JoinExt` and be able to use all the other goodies in there.


mlange-42:
I think the decision to switch to `display_utils` depends on whether we can also use its functionality to our advantage elsewhere, which we probably can.

mlange-42 (approved):

adamreichold:
> I think the decision to switch to `display_utils` depends on whether we can also use its functionality to our advantage elsewhere, which we probably can.

Will open a follow-up PR. Do you have a specific place in mind?

mlange-42:
> Do you have a specific place in mind?

No not really. Would have to scan the code, probably the most promising place would be `print.rs`. But in case there are no further places where `display_utils` significantly simplifies the code or avoids allocations, I would vote for not including the additional dependency.

PR #178, opened 2021-05-13
Replace bespoke lazy join by dependency on display_utils crate.

adamreichold:
On a cursory look, I did not find another place we I'd jump for `display_utils` though...


adamreichold:
@mlange-42 Do we want to pull in the dependency?

mlange-42:
Sorry for the delay, this got lost somehow. As we still use the new dependency only for the joins, but no other of its functionality, I am a bit undecided.

adamreichold:
If we do not use them, the changes to `print.rs` should probably be applied using our own lazy join.

adamreichold:
#180 would be that alternative if we do not want a dependency on `display_utils`.

PR #180, opened 2021-05-29
Add further uses of lazy join when printing.

adamreichold:


mlange-42:
Would prefer this over #178.

mlange-42 (approved):

